PR #2, opened 2015-10-22
WIP: transformers that modify y

GaelVaroquaux:
This is almost ready for early discussion:

Best way to look at the proposal is:
https://github.com/GaelVaroquaux/enhancement_proposals/blob/transform_y/slep001/discussion.rst

### DIFF
--- a/slep001/discussion.rst
+++ b/slep001/discussion.rst
@@ -10,7 +10,11 @@ Transformers that modify their target
         X_transform = estimator.transform(X)
         estimator.fit(X, y=None).transform(X) == estimator.fit_transform(X, y)
 
-    Many usecases require modifying y. How do we support this?
+    Within a chain or processing sequence of estimators, many usecases
+    require modifying y. How do we support this?
+    
+    Doing many of these things is possible "by hand". The question is:
+    how to avoid writing custom connecting logic.
 
 .. sectnum::
 
@@ -137,14 +141,24 @@ conceptual difficulty
    "data processing pipeline", but we use "heap" to avoid confusion with
    the pipeline object.
 
-   Stacks combining many steps of pipelines and meta-estimators become
+   Heaps combining many steps of pipelines and meta-estimators become
    very hard to inspect and manipulate, both for the user, and for
    pipeline-management (aka "heap-management") code. Currently, these
    difficulties are mostly in user code, so we don't see them too much in
    scikit-learn. Here are concrete examples
 
-   #. Trying to retrieve coefficients from a models estimated in a
-      "heap". Solving this problem requires 
+   #. Trying to retrieve coefficients from a model estimated in a
+      "heap". Eg: 
+      
+      * you know there is a lasso in your stack and you want to
+        get it's coef (in whatever space that resides?):
+        `pipeline.named_steps['lasso'].coef_` is possible.
+
+      * you want to retrieve the coef of the last step:
+        `pipeline.steps[-1][1].coef_` is possible.
+
+      With meta estimators this is tricky.
+      Solving this problem requires 
       https://github.com/scikit-learn/scikit-learn/issues/2562#issuecomment-27543186
       (this enhancement proposal is not advocating to solve the problem
       above, but pointing it out as an illustration)
@@ -154,19 +168,22 @@ conceptual difficulty
       that there was mostly one object to modify to do the dispatching,
       the Pipeline object.
 
-   #. A future, out-of-core "conductor" object to fit a "stack" in out of
+   #. A future, out-of-core "conductor" object to fit a "heap" in out of
       core by connecting it to a data-store would need to have a
-      representation of the stack. For instance, when chaining random
+      representation of the heap. For instance, when chaining random
       projections with Birch coresets and finally SGD, the user would
       need to specify that random projections are stateless, birch needs
       to do one pass of the data, and SGD a few. Given this information,
       the conductor could orchestrate pull the data from the data source,
       and sending it to the various steps. Such an object is much harder
       to implement if the various steps are to be combined in a heap.
+      Note that the scikit-learn pipeline can only implement a linear
+      "chain" like set of processing. For instance a One vs All will
+      never be able to be implemented in a scikit-learn pipeline.
 
-      Note that this is not a problem in non out-of-core settings, in the
-      sense that the BirchCoreSet meta-estimator would take care of doing
-      a pass on the data before feeding it to its sub estimator.
+      This is not a problem in non out-of-core settings, in the sense
+      that the BirchCoreSet meta-estimator would take care of doing a
+      pass on the data before feeding it to its sub estimator.
 
 In conclusion, meta-estimators are harder to comprehend (problem 1) and
 write (problem 2).
@@ -183,9 +200,9 @@ Option B: transformer-like that modify y
     1. Changing the semantics of transformers to modify y and return
        something more complex than a data matrix X
 
-    2. Introducing a new type of object
+    2. Introducing new methods (and a new type of object)
 
-    Their is an emerging consensus for option 2.
+    There is an emerging consensus for option 2.
 
 Proposal
 .........
@@ -229,13 +246,22 @@ Design questions
    For some usecases, test time needs to modify the number of samples
    (for instance data loading from a file). However, these will by
    construction a problem for eg cross-val-score, as they need to
-   generate a y_true. It is thus unclear that the data-loading usecases
-   can be fully integrated in the CV framework (which is not an argument
-   against enabling them).
+   generate a y_true. Indeed, the problem is the following:
 
-   For our CV framework, we need the number of samples to remain constant
-   (to have correspondence between y_pred and _true). This is an argument
-   for:
+   - To measure an error, we need y_true at the level of
+     'cross_val_score' or GridSearch
+ 
+   - y_true is created inside the pipeline by the data-loading object.
+
+   It is thus unclear that the data-loading usecases can be fully
+   integrated in the CV framework (which is not an argument against
+   enabling them).
+
+   |
+
+   For our CV framework, we need the number of samples to remain
+   constant: for each y_pred, we need a corresponding y_true. This is an
+   argument for:
     
    #. Accepting both transform and transform_pipe
 
@@ -270,20 +296,21 @@ In term of name choice, the rational would be to have method names that
 are close to 'fit' and 'transform', to make discoverability and
 readability of the code easier.
 
-* Name of the object:
+* Name of the object (referred in the docs):
   - TransformPipe
   - PipeTransformer
-  - FilterTransform
+  - TransModifier
 
 * Method to fit and apply on training 
   - fit_pipe
   - pipe_fit
   - fit_filter
+  - fit_modify
 
 * Method to apply on new data (not always available)
   - transform_pipe
   - pipe_transform
-  - transform_filter
+  - trans_modify
 
 Benefits
 .........
@@ -301,7 +328,7 @@ Benefits
 Limitations
 ............
 
-* Introducing new methods, and a new type of estimator object:. There are
+* Introducing new methods, and a new type of estimator object. There are
   probably a total of **3 new methods** that will get introduced by this
   enhancement: fit_pipe, transform_pipe, and partial_fit_pipe
 
--- a/slep001/discussion.rst
+++ b/slep001/discussion.rst
@@ -211,7 +211,7 @@ Option B: transformer-like that modify y
    
    Do we modify all transformers to return a y (y=None for unsupervised
    transformers that are not given y?). This sounds like leading to code
-   full of surprised and difficult to maintain from the user perspective.
+   full of surprises and difficult to maintain from the user perspective.
 
    We would loose the contract that the number of samples is unchanged by
    a transformer. This contract is very useful (eg for model selection:
--- a/slep001/discussion.rst
+++ b/slep001/discussion.rst
@@ -1,0 +1,295 @@
+=====================================
+Transformers that modify their target
+=====================================
+
+.. topic:: **Summary**
+
+    Transformers implement::
+
+        self = estimator.fit(X, y=None)
+        X_transform = estimator.transform(X)
+        estimator.fit(X, y=None).transform(X) == estimator.fit_transform(X, y)
+
+    Many usecases require modifying y. How do we support this?
+
+.. sectnum::
+
+.. contents:: Table of contents
+   :depth: 2
+
+Rational
+==========
+
+Summary of the contract of transformers
+----------------------------------------
+
+* .transform(...) returns a data matrix X
+
+* .transform(...) returns one feature vector for each sample of the input
+
+* .fit_transform(...) is the same and .fit(...).transform(...)
+
+Examples of usecases targetted
+-------------------------------
+
+#. Over sampling:
+
+   #. Class rembalancing: over sampling the minority class in
+      unbalanced dataset
+   #. Data enhancement (nudgging images for instance)
+
+#. Under-sampling
+
+   #. Stateless undersampling: Take one sample out of two
+   #. Stateful undersampling: apply clustering and transform to cluster
+      centers
+   #. Coresets: return a smaller number of samples and associated sample
+      weights
+
+#. Outlier detection:
+
+   #. Remove outlier from train set
+   #. Create a special class 'y' for outliers
+
+#. Completing y:
+
+   #. Missing data imputation on y
+   #. Semi-supervised learning (related to above)
+
+#. Data loading / conversion
+
+   #. Pandas in => (X, y) out
+   #. Filename in => (X, y) with multiple samples (very useful in
+      combination with online learning)
+   #. Database query => (X, y) out
+
+#. Aggregate statistics over multiple samples
+
+   #. Windowing-like functions on time-series
+  
+   In a sense, these are dodgy with scikit-learn's cross-validation API
+   that knows nothing about sample structure. But the refactor of the CV
+   API is really helping in this regard.
+
+____
+
+These usecases pretty much require breaking the contract of the
+Transformer, as detailed above.
+
+The intuition driving this enhancement proposal is that the more the
+data-processing pipeline becomes rich, the more the data grow, the more
+the usecases above become important.
+
+Enhancements proposed
+=======================
+
+Option A: meta-estimators
+---------------------------
+
+Proposal
+........
+
+This option advocates that any transformer-like usecase that wants to
+modify y or the number of samples should not be a transformer-like but a
+specific meta-estimator. A core-set object would thus look like:
+
+* From the user perspective::
+
+     from sklearn.sample_shrink import BirchCoreSet
+     from sklearn.ensemble import RandomForest
+     estimator = BirchCoreSet(RandomForest())
+
+* From the developer perspective::
+
+     class BirchCoreSet(BaseEstimator):
+
+        def fit(self, X, y):
+            # The logic here is wrong, as we need to handle y:
+            super(BirchCoreSet, self).fit(X)
+            X_red = self.subcluster_centers_
+            self.estimator_.fit(X_red)
+
+Benefits
+.........
+
+#. No change to the existing API
+
+#. The meta-estimator pattern is very powerful, and pretty much anything
+   is possible.
+
+Limitations
+............
+
+The different limitations listed below are variants of the same
+conceptual difficulty
+
+#. It is hard to have mental models and garantees of what a
+   meta-estimator does, as it is by definition super versatile
+
+   This is both a problem for the beginner, that needs to learn them on
+   an almost case-by-case basis, and for the advanced user, that needs to
+   maintain a set of case-specific code
+
+#. The "estimator heap" problem. 
+
+   Here the word heap is used to denote the multiple pipelines and
+   meta-estimators. It corresponds to what we would naturally call a
+   "data processing pipeline", but we use "heap" to avoid confusion with
+   the pipeline object.
+
+   Stacks combining many steps of pipelines and meta-estimators become
+   very hard to inspect and manipulate, both for the user, and for
+   pipeline-management (aka "heap-management") code. Currently, these
+   difficulties are mostly in user code, so we don't see them too much in
+   scikit-learn. Here are concrete examples
+
+   #. Trying to retrieve coefficients from a models estimated in a
+      "heap". Solving this problem requires 
+      https://github.com/scikit-learn/scikit-learn/issues/2562#issuecomment-27543186
+      (this enhancement proposal is not advocating to solve the problem
+      above, but pointing it out as an illustration)
+
+   #. DaskLearn has modified the logic of pipeline to expose it as a
+      computation graph. The reason that it was relatively easy to do is
+      that there was mostly one object to modify to do the dispatching,
+      the Pipeline object.
+
+   #. A future, out-of-core "conductor" object to fit a "stack" in out of
+      core by connecting it to a data-store would need to have a
+      representation of the stack. For instance, when chaining random
+      projections with Birch coresets and finally SGD, the user would
+      need to specify that random projections are stateless, birch needs
+      to do one pass of the data, and SGD a few. Given this information,
+      the conductor could orchestrate pull the data from the data source,
+      and sending it to the various steps. Such an object is much harder
+      to implement if the various steps are to be combined in a heap.
+
+      Note that this is not a problem in non out-of-core settings, in the
+      sense that the BirchCoreSet meta-estimator would take care of doing
+      a pass on the data before feeding it to its sub estimator.
+
+In conclusion, meta-estimators are harder to comprehend (problem 1) and
+write (problem 2).
+
+That said, we will never get rid of meta estimators. It is a very
+powerful pattern. The discussion here is about extending a bit the
+estimator API to have a less pressing need for meta-estimators.
+
+Option B: transformer-like that modify y
+------------------------------------------
+
+.. note:: Two variants of this option exist:
+
+    1. Changing the semantics of transformers to modify y and return
+       something more complex than a data matrix X
+
+    2. Introducing a new type of object
+
+    Their is an emerging consensus for option 2.
+
+Proposal
+.........
+
+Introduce a `TransformPipe` type of object with the following API
+(names are discussed below):
+
+* `X_new, y_new = estimator.fit_pipe(X, y)`
+
+* `X_new, y_new = estimator.transform_pipe(X, y)`
+
+Or:
+
+* `X_new, y_new, sample_props = estimator.fit_pipe(X, y)`
+
+* `X_new, y_new, sample_props = estimator.transform_pipe(X, y)`
+
+Contracts (these are weaker contracts than the transformer:
+
+* Neither `fit_pipe` nor `transform_pipe` are guarantied to keep the
+  number of samples unchanged.
+
+* transform_pipe is not equivalent to .fit_pipe.transform
+
+Design questions
+....................
+
+#. Should there be a fit method?
+
+   In such estimators, it may not be a good idea to call fit rather than
+   fit_pipe (for instance in coreset).
+
+
+#. At test time, how does a pipeline use such an object?
+
+   #. Should there be a transform method used at test time?
+
+   #. What to do with objects that implement both `transform` and
+      `transform_pipe`?
+
+   For some usecases, test time needs to modify the number of samples
+   (for instance data loading from a file). However, these will by
+   construction a problem for eg cross-val-score, as they need to
+   generate a y_true. It is thus unclear that the data-loading usecases
+   can be fully integrated in the CV framework (which is not an argument
+   against enabling them).
+
+   For our CV framework, we need the number of samples to remain constant
+   (to have correspondence between y_pred and _true). This is an argument
+   for:
+    
+   #. Accepting both transform and transform_pipe
+
+   #. Having the pipeline 'predict' use 'transform' on its
+      intermediate steps
+   
+#. How do we deal with sample weights and other sample properties
+
+   This discussion feeds in the `sample_props` discussion (that should
+   be discussed in a different enhancement proposal).
+
+   The suggestion is to have the sample properties as a dictionary of
+   arrays `sample_props`.
+
+   **Example usecase** useful to think about sample properties: coresets:
+   given (X, y) return (X_new, y_new, weights) with a much smaller number
+   of samples.
+
+   This example is interesting because it shows that PipeTransforms can
+   legitimately create sample properties.
+
+   **Proposed solution**:
+
+   * PipeTransforms always return (X_new, y_new, sample_props) where
+     sample_props can be an empty dictionary.
+
+
+Naming suggestions
+..................
+
+In term of name choice, the rational would be to have method names that
+are close to 'fit' and 'transform', to make discoverability and
+readability of the code easier.
+
+* Name of the object:
+  - TransformPipe
+  - PipeTransformer
+  - FilterTransform
+
+* Method to fit and apply on training 
+  - fit_pipe
+  - pipe_fit
+  - fit_filter
+
+* Method to apply on new data (not alway available)
+  - transform_pipe
+  - pipe_transform
+  - transform_filter
+
+Benefits
+.........
+
+Limitations
+............
+
+
+
--- a/slep001/discussion.rst
+++ b/slep001/discussion.rst
@@ -223,90 +223,109 @@ Option B: transformer-like that modify y
 Proposal
 .........
 
-Introduce a `TransformPipe` type of object with the following API
+Introduce a `TransModifier` type of object with the following API
 (names are discussed below):
 
-* `X_new, y_new = estimator.fit_pipe(X, y)`
+* `X_new, y_new = estimator.fit_modify(X, y)`
 
-* `X_new, y_new = estimator.transform_pipe(X, y)`
+* `X_new, y_new = estimator.trans_modify(X, y)`
 
 Or:
 
-* `X_new, y_new, sample_props = estimator.fit_pipe(X, y)`
+* `X_new, y_new, sample_props = estimator.fit_modify(X, y)`
 
-* `X_new, y_new, sample_props = estimator.transform_pipe(X, y)`
+* `X_new, y_new, sample_props = estimator.trans_modify(X, y)`
 
 Contracts (these are weaker contracts than the transformer:
 
-* Neither `fit_pipe` nor `transform_pipe` are guarantied to keep the
+* Neither `fit_modify` nor `trans_modify` are guarantied to keep the
   number of samples unchanged.
 
-* transform_pipe is not equivalent to .fit_pipe.transform
+* `fit_modify` may not exist (questionnable)
 
-Design questions
-....................
+Design questions and difficulties
+..................................
 
-#. Should there be a fit method?
+Should there be a fit method?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   In such estimators, it may not be a good idea to call fit rather than
-   fit_pipe (for instance in coreset).
+In such estimators, it may not be a good idea to call fit rather than
+fit_modify (for instance in coreset).
 
 
-#. At test time, how does a pipeline use such an object?
+How does a pipeline use such an object?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   #. Should there be a transform method used at test time?
+In particular at test time?
 
-   #. What to do with objects that implement both `transform` and
-      `transform_pipe`?
+#. Should there be a transform method used at test time?
 
-   For some usecases, test time needs to modify the number of samples
-   (for instance data loading from a file). However, these will by
-   construction a problem for eg cross-val-score, as they need to
-   generate a y_true. Indeed, the problem is the following:
+#. What to do with objects that implement both `transform` and
+   `trans_modify`?
 
-   - To measure an error, we need y_true at the level of
-     'cross_val_score' or GridSearch
+**Creating y in a pipeline makes error measurement harder** For some
+usecases, test time needs to modify the number of samples (for instance
+data loading from a file). However, these will by construction a problem
+for eg cross-val-score, as in supervised settings, these expect a y_true.
+Indeed, the problem is the following:
+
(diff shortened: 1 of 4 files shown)


amueller:
I think you should state more clearly that this is for within a chain or processing sequence of estimators. Doing many of these things is possible  "by hand". the point is that you don't want to write custom connecting logic.

amueller:
maybe "chain" instead of heap and stack?

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+conceptual difficulty
+
+#. It is hard to have mental models and garantees of what a
+   meta-estimator does, as it is by definition super versatile
+
+   This is both a problem for the beginner, that needs to learn them on
+   an almost case-by-case basis, and for the advanced user, that needs to
+   maintain a set of case-specific code
+
+#. The "estimator heap" problem.
```
amueller: maybe "chain" instead of heap and stack?

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+   "data processing pipeline", but we use "heap" to avoid confusion with
+   the pipeline object.
+
+   Stacks combining many steps of pipelines and meta-estimators become
+   very hard to inspect and manipulate, both for the user, and for
+   pipeline-management (aka "heap-management") code. Currently, these
+   difficulties are mostly in user code, so we don't see them too much in
+   scikit-learn. Here are concrete examples
+
+   #. Trying to retrieve coefficients from a models estimated in a
```
amueller: I think the point would be made stronger by a working example. You know there is a ``lasso`` in your stack and you want to get it's ``coef_`` [in whatever space that resides?].
```python
pipeline.named_steps['lasso'].coef_
```
is possible.
With a chain of meta estimators this is tricky.

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+      https://github.com/scikit-learn/scikit-learn/issues/2562#issuecomment-27543186
+      (this enhancement proposal is not advocating to solve the problem
+      above, but pointing it out as an illustration)
+
+   #. DaskLearn has modified the logic of pipeline to expose it as a
+      computation graph. The reason that it was relatively easy to do is
+      that there was mostly one object to modify to do the dispatching,
+      the Pipeline object.
+
+   #. A future, out-of-core "conductor" object to fit a "stack" in out of
```
amueller: I still maintain that having the conductor object try to handle this makes it much more complicated. I walked through what is necessary to do this with a meta-estimator approach, but I'm happy to show pseudo code.

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+      the conductor could orchestrate pull the data from the data source,
+      and sending it to the various steps. Such an object is much harder
+      to implement if the various steps are to be combined in a heap.
+
+      Note that this is not a problem in non out-of-core settings, in the
+      sense that the BirchCoreSet meta-estimator would take care of doing
+      a pass on the data before feeding it to its sub estimator.
+
+In conclusion, meta-estimators are harder to comprehend (problem 1) and
+write (problem 2).
```
amueller: I would say "interface with". I think they are much easier to write.

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+That said, we will never get rid of meta estimators. It is a very
+powerful pattern. The discussion here is about extending a bit the
+estimator API to have a less pressing need for meta-estimators.
+
+Option B: transformer-like that modify y
+------------------------------------------
+
+.. note:: Two variants of this option exist:
+
+    1. Changing the semantics of transformers to modify y and return
```
amueller: I think there is 1.a) add a new method to transformers and 1.b) replace transform.

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+
+* transform_pipe is not equivalent to .fit_pipe.transform
+
+Design questions
+....................
+
+#. Should there be a fit method?
+
+   In such estimators, it may not be a good idea to call fit rather than
+   fit_pipe (for instance in coreset).
```
amueller: or subsampling, which is sort of an easier concept?

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+#. At test time, how does a pipeline use such an object?
+
+   #. Should there be a transform method used at test time?
+
+   #. What to do with objects that implement both `transform` and
+      `transform_pipe`?
+
+   For some usecases, test time needs to modify the number of samples
+   (for instance data loading from a file). However, these will by
+   construction a problem for eg cross-val-score, as they need to
```
amueller: I don't understand the problem.

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+   generate a y_true. It is thus unclear that the data-loading usecases
+   can be fully integrated in the CV framework (which is not an argument
+   against enabling them).
+
+   For our CV framework, we need the number of samples to remain constant
```
amueller: I don't follow.

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+
+
+Naming suggestions
+..................
+
+In term of name choice, the rational would be to have method names that
+are close to 'fit' and 'transform', to make discoverability and
+readability of the code easier.
+
+* Name of the object:
```
amueller: Is this the name for the mixin or just how we refer to them in the docs? It's good to have a name so we can talk about them.

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+  - TransformPipe
+  - PipeTransformer
+  - FilterTransform
```
amueller: filter suggests downsampling. -1

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+
+* Method to fit and apply on training
```
amueller: ``fit_process`` and ``process``?
if we don't have a ``fit`` function, we actually don't need to worry about breach of contract stuff.

slep001/discussion.rst
```diff
@@ -0,0 +1,295 @@
...
+* `X_new, y_new, sample_props = estimator.fit_pipe(X, y)`
+
+* `X_new, y_new, sample_props = estimator.transform_pipe(X, y)`
+
+Contracts (these are weaker contracts than the transformer:
+
+* Neither `fit_pipe` nor `transform_pipe` are guarantied to keep the
+  number of samples unchanged.
+
+* transform_pipe is not equivalent to .fit_pipe.transform
```
amueller: not sure if that was intended: ``fit_pipe`` won't return self, so you can't chain in that way.

slep001/discussion.rst
amueller: On seconds though, users might expect these to do the same thing:
```python

thing = Thing()
X_1, y_1, sample_props_1 = thing.fit_pipe(X, y, sample_props)
X_2, y_2, sample_props_2 = thing.transform_pipe(X, y, sample_props)
```
And we should choose naming such that this is not suggested.

slep001/discussion.rst
```diff
@@ -0,0 +1,311 @@
...
+Introduce a `TransformPipe` type of object with the following API
+(names are discussed below):
+
+* `X_new, y_new = estimator.fit_pipe(X, y)`
+
+* `X_new, y_new = estimator.transform_pipe(X, y)`
+
+Or:
+
+* `X_new, y_new, sample_props = estimator.fit_pipe(X, y)`
```
amueller: this is actually necessary for supporting even undersampling or oversampling. Without this, use cases are limited to not using sample weights.
[which is one more argument for keeping "transform" around. And a reason why I probably wouldn't teach the new method first ;)]

slep001/discussion.rst
amueller: thesaurus also suggests "alter" "handle" "treat" "transmute" "compute" "run" "propagate" "mutate" "transmogrify" "transfigure" and "replace" [that one is actually interesting!] (also "convert" which I don't like)

slep001/discussion.rst
```diff
@@ -0,0 +1,311 @@
...
+
+Naming suggestions
+..................
+
+In term of name choice, the rational would be to have method names that
+are close to 'fit' and 'transform', to make discoverability and
+readability of the code easier.
+
+* Name of the object:
+  - TransformPipe
```
amueller: Tube? Chute? Hose? ^^

slep001/discussion.rst
```diff
@@ -0,0 +1,311 @@
...
+* The API being more versatile, it will be easier to create
+  application-specific code or framework wrappers (ala DaskLearn) that
+  are scikit-learn compatible, and thus that can be used with the
+  parameter-selection framework. This will be especially true for ETL
+  (extract transform and load) pattern.
+
+Limitations
+............
+
+* Introducing new methods, and a new type of estimator object:. There are
```
amueller: either get rid of the dot or the colon?

slep001/discussion.rst
```diff
@@ -0,0 +1,311 @@
...
+* Method to apply on new data (not always available)
+  - transform_pipe
+  - pipe_transform
+  - transform_filter
+
+Benefits
+.........
+
+* Many usecases listed above will be implemented scikit-learn without a
+  meta-estimator, and thus will be easy to use (eg in a pipeline). Many
```
amueller: Easier to interface with? I guess easier to use using GridSearchCV is true...

amueller:
Just a note: misses the part that @agramfort wanted to write,  user story code.
@agramfort said he already has ugly solutions that would be nicer with the new interface. It would be good to see them.

It would probably be easy to come up with some outlier or resampling examples, but more "real live" ones would be cool.

amueller:
overall looks great and I agree with the summary mostly. the "should this be a new class" discussion is touched upon only briefly, and the "should the number of return arguments vary" discussion, too.

amueller:
ping @jnothman [no rush] @glouppe

slep001/discussion.rst
amueller: Processor? Transmuter? Mutator? Transfiguerer?

slep001/discussion.rst
GaelVaroquaux: "chain" is linear. But point taken, I am rephrasing a bit this paragraph.

GaelVaroquaux:
> I think you should state more clearly that this is for within a chain or
> processing sequence of estimators. Doing many of these things is possible "by
> hand". the point is that you don't want to write custom connecting logic.

Thanks. Addressed.

slep001/discussion.rst
GaelVaroquaux: > I think the point would be made stronger by a working example. You know
> there is a lasso in your stack and you want to get it's coef_ [in
> whatever space that resides?].

Addressed.

slep001/discussion.rst
GaelVaroquaux: > I still maintain that having the conductor object try to handle this makes it
> much more complicated.

Agreed. But I think that this proposal will make it simpler.

> I walked through what is necessary to do this with a meta-estimator
> approach, but I'm happy to show pseudo code.

Do you want to put this up in a gist? It would be useful.

slep001/discussion.rst
GaelVaroquaux: > > +In conclusion, meta-estimators are harder to comprehend (problem 1) and
> > +write (problem 2).

> I would say "interface with". I think they are much easier to write.

Not clear. They need to use the delegate decorators. Most people in
machine learning have no idea what delegation is, or what "delegate"
means.

slep001/discussion.rst
GaelVaroquaux: > I think there is 1.a) add a new method to transformers and 1.b) replace transform.

Well, given that we are pretty lax about inheritence, and given that the
mandatory / optional methods of the new interface is still not figured
out, I think that 1.a) is similar enough to 2. I have edited the text to
make it open to 1.a).

slep001/discussion.rst
GaelVaroquaux: > or subsampling, which is sort of an easier concept?

I like coreset because they raise a fair amount of problem, like sample
weights, but also the fact that they need fitting (subsampling might
not).

slep001/discussion.rst
GaelVaroquaux: > I don't understand the problem.

I tried to make things clearer.

slep001/discussion.rst
GaelVaroquaux: > I don't follow.

I tried to make things clearer.

slep001/discussion.rst
GaelVaroquaux: > Is this the name for the mixin or just how we refer to them in the docs? It's
> good to have a name so we can talk about them.

Let's start by finding a good name for the docs.

slep001/discussion.rst
GaelVaroquaux: > filter suggests downsampling. -1

Yes, I agree with you.

slep001/discussion.rst
GaelVaroquaux: > fit_process and process?

I would like the 'transform-like' method to contain the word transform
(or at least 'transsomething') for discoverability.

> if we don't have a fit function, we actually don't need to worry about breach
> of contract stuff.

Indeed.

slep001/discussion.rst
GaelVaroquaux: > > +* transform_pipe is not equivalent to .fit_pipe.transform

> not sure if that was intended: fit_pipe won't return self, so you can't chain
> in that way.

Yes. It was more conceptual than practical.

slep001/discussion.rst
GaelVaroquaux: > On seconds though, users might expect these to do the same thing:

> thing = Thing()
> X_1, y_1, sample_props_1 = thing.fit_pipe(X, y, sample_props)
> X_2, y_2, sample_props_2 = thing.transform_pipe(X, y, sample_props)

> And we should choose naming such that this is not suggested.

I am not sure I am following what you are saying above.

I've been thinking about name. I think that the prefix that captures the
idea is 'trans'. It has a sexual connotation that we have to be careful
about. But maybe there is something to be found here, around it.

brainstorming mode:

* find a new name for the "transform-like method" around this idea:

  translate, transmutate, transcript, transfix, transcend, transfer,
  transmod

  for the 'fit like': the same with 'fit_' in front.

* fit_modify / trans_modify.

I somewhat like this last suggestion. I find that 'TransModifier' could be
an object name. I also find that it convey this idea that it is more
violent than 'transform' but in the same alley.

slep001/discussion.rst
GaelVaroquaux: > this is actually necessary for supporting even undersampling or oversampling.
> Without this, use cases are limited to not using sample weights.

Agreed.

> [which is one more argument for keeping "transform" around. And a reason why I
> probably wouldn't teach the new method first ;)]

Agreed.

In terms of 'teachability' and 'discovarability', my focus is more on
have a good vocabulary by now.

slep001/discussion.rst
GaelVaroquaux: > > +* Many usecases listed above will be implemented scikit-learn without a
> > +  meta-estimator, and thus will be easy to use (eg in a pipeline). Many

> Easier to interface with? I guess easier to use using GridSearchCV is true...

If pipeline hides this connections away from the user (provided that we
can come up with the right semantics), it should be easier to use: just a
question of make_pipeline.

GaelVaroquaux:
> Just a note: misses the part that @agramfort wanted to write, user story code.
> @agramfort said he already has ugly solutions that would be nicer with the new
> interface. It would be good to see them.

ping @agramfort

slep001/discussion.rst
GaelVaroquaux: > Processor? Transmuter? Mutator? Transfiguerer?

TransModifier
TransModifier.fit_modify
TransModifier.trans_modify

My favorit. I can (almost) see myself teaching this.

GaelVaroquaux:
I just realized that this proposal is adding a conceptual complexity (and thus burden for the user):

We will now have 3 methods that can modify X and are legitimate in a pipeline (interim names for now):

- fit_pipe
- transform_pipe
- transform

The question is: at fit time, in a pipeline, what has the priority on what, if not only one of these methods is present. In my head it is pretty clear (it's in the order listed above). But it might be confusing for the user.


We need fit_pipe and transform_pipe to exist both (and be different) to tackle things such as outlier detection. However, in most cases, we probably do not need transform_pipe

podshumok:
Hello!

I used to deal with such things too.

My current approach (which should work with stable sklearn) is to add some ABCs and monkey patch `Pipeline` to use them.

ABC is abstract base class (it has hooks for `isinstance` calls)

```python
>>> lab_enc = LabelEncoder()
...
>>> isinstance(lab_enc, ABC_Y_Transformer)
True
```

Pipeline is patched so that if estimator `step` passes the `isinstance` check with `ABC_Y_Transformer`, then `fit` and `transform` methods of `step` will be called with `Y` instead of `X`.

My current solution for such ABC: hardcode all sklearn (and my own) classes in `ABC_Y_Transformer..__instancecheck__` so that `isinstance(obj, ABC_Y_Transformer)` call becames `obj.__class__ in predefined_set_of_classes`.

This way I can put label transformers and `X` transformers in a single pipeline. 

The same logic can be applied to other cases, where `X` and `Y` should be modified together.

Instead of this hacky ABC, sklern can have two additional mixins - `LabelTransformerMixin` and `XYTransformerMixin`.  And pipeline could be patched the same way as with ABC.

This way, modifications to sklearn code will be small, new version of sklearn will be (almost) compatible with previous ones and overall API changes are small.

Modifications needed: add mixins, add parent to current label transformers, make pipeline respect new mixins.

mblondel:
Could you elaborate the issues with the current design and why we can't transform y at the moment? We do have LabelBinarizer, although it's never used in a pipeline.

agramfort:
the code I'd like to write it:

```
est = make_pipeline(ClassBalancer(), SVC())  # balance classes by subsampling before calling SVC
est = make_pipeline(DataAugmentator(), RandomForest())  #  data augmentation followed by RF
est = make_pipeline(BirchCoresets(), GMM())  # coresets and GMM for unsurpervised
est = make_pipeline(BirchCoresets(), Classifier())  # coresets and then surpervised than could do partial_fit if Classifier has it
```

then for out of core / file based usage:

```
est = make_pipeline(FilesToXy(), Classifier())
```

FilesToXy could do with partial or full "fit" conversation from files to X, y pairs eventually with sample props.

Amend if you see more.

GaelVaroquaux:
> Instead of this hacky ABC, sklern can have two additional mixins -
> LabelTransformerMixin and XYTransformerMixin.

What you are suggesting goes in the same direction of option B. However
we have a design policy which is to try to rely as much as possible on
interfaces rather than inheritance (or typing). The reason is that it
make things more explicit: given the name of a method (which I see from
the code I am typing or reading) I should be able to infer what a class
does. Thus the need to introduce new methods.

GaelVaroquaux:
> FilesToXy could do with partial or full "fit" conversation from files to X, y
> pairs eventually with sample props.

Can you read my discussion (part 2.2.2, bullet 2) on why I think this is
hard / impossible to integrate in model selection framework, and tell me
if you understand / agree.

GaelVaroquaux:
I have added a note on these issue at the beginning of opion B. Tell me
if it is clear.

On Thu, Oct 22, 2015 at 07:54:12PM -0700, Mathieu Blondel wrote:
> Could you elaborate the issues with the current design and why we can't
> transform y at the moment? We do have LabelBinarizer, although it's never used
> in a pipeline.

> —
> Reply to this email directly or view it on GitHub.*


-- 
    Gael Varoquaux
    Researcher, INRIA Parietal
    NeuroSpin/CEA Saclay , Bat 145, 91191 Gif-sur-Yvette France
    Phone:  ++ 33-1-69-08-79-68
    http://gael-varoquaux.info            http://twitter.com/GaelVaroquaux

ogrisel:
I think we need to write down how we would implement some example pipelines using the proposed API:


- pipeline 1: [
      - CSV file collection parser,  # stateless (does not need fit)
      - under sampling of the majority class based on the extracted y,  # stateful and stochastic at fit time
      - one hot encoder for categorical variable columns,  # stateful (need one pass)
      - max abs scaler,  # stateful (need one pass)
      - logistic regression (possibly out of core if implemented with SGDClassifier)  # stateful (can need several passes)
]

- pipeline 2 that compresses the data:
    - BirchCoreSet,  # reduce the number of sample after one pass over the data
    - SVC

Note: pipeline 1 should ideally support batch and out-of-core fitting. In the out-of-core case I am afraid that the behavior of the stochastic behavior of the under sampler need to be aware it it's transforming data for fitting downstream estimators (in which case it need to actually perform under sampling) or if all the downstream estimators have been fitted and in which case it behaves like a pass-through (identity function) to make actual predictions.

Maybe the pipeline instance could be 'smart' and understand that it needs to disable the under sampler at prediction time. However we might face the same problem with other models such as a jitter / nudger used for sample augmentation that is only meaningful at fit time.

slep001/discussion.rst
amueller: The meta-estimators I had in mind are always linear, and the new objects are always linear. ``FeatureUnion`` is the only thing that can make it a DAG, right?

slep001/discussion.rst
amueller: On 10/23/2015 12:38 AM, Gael Varoquaux wrote:
>
> > +In conclusion, meta-estimators are harder to comprehend (problem 1) 
> and > +write (problem 2). I would say "interface with". I think they 
> are much easier to write.
> Not clear. They need to use the delegate decorators. Most people in 
> machine learning have no idea what delegation is, or what "delegate" 
> means.
>
Point taken.

slep001/discussion.rst
amueller: ooohhh. for some reason  I thought that worked. We can make ``Pipeline.score`` work. But using a custom scorer it's a bit unclear.

amueller:
> We will now have 3 methods that can modify X and are legitimate in a pipeline (interim names for now):

    fit_pipe
    transform_pipe
    transform

The question is: at fit time, in a pipeline, what has the priority on what, if not only one of these methods is present. In my head it is pretty clear (it's in the order listed above). But it might be confusing for the user.

I think we should not have objects that have both ``transform`` and  ``transform_pipe``. And I think every method that has ``fit_pipe`` also has ``transform_pipe``.
So an object either has ``fit`` and ``transform`` or ``fit_pipe`` and ``transform_pipe``. Then the semantics are pretty clear to me. ``fit_pipe`` is called during ``fit`` and ``transform_pipe`` is called during ``predict``, ``transform`` and ``score``.

amueller:
Fun question: does the pipeline have ``fit_pipe`` and ``transform_pipe`` ?

amueller:
@podshumok we want to modify both ``X`` and ``y`` in the pipeline.

amueller:
@ogrisel I feel like we should leave ``partial_fit`` out of this. I think it is pretty unrelated. I think the solution for ``partial_fit`` will not depend on whether the pipeline calls ``transform`` or ``transform_pipe``

kastnerkyle:
There was a pretty serious discussion on reddit about this type of stuff (and tying in multi-core) [a few days ago](https://www.reddit.com/r/MachineLearning/comments/3pgx6y/dask_sklearn_experiment_reuse_intermediate/) . Just a pointer for reference/future, a sample of opinions about what issues might be important to users.

kastnerkyle:
as far as objects not having both ``transform`` and ``transform_pipe`` - how do we handle legacy support e.g. PCA? Or does you mean just ignoring ``transform`` on the objects inside this new OOC pipeline thing

slep001/discussion.rst
kastnerkyle: The difference between ``fit`` time and ``predict`` time in the case of things like the previously discussed undersampler effectively create a conditional DAG, depending on whether you are at ``fit`` or ``predict`` time

podshumok:
>What you are suggesting goes in the same direction of option B. However
>we have a design policy which is to try to rely as much as possible on
>interfaces rather than inheritance (or typing).

then maybe it is better to have method indicating whatever we transform data, labels, or both. instead of adding multiple new methods for actual fitting and/or transforming.

>@podshumok we want to modify both X and y in the pipeline.
yep, I covered this point. The reason I focused on Y's only is that such transformers are already in sklearn, but they are not usable with pipelines.

slep001/discussion.rst
GaelVaroquaux: > ooohhh. for some reason I thought that worked. We can make
> Pipeline.score work. But using a custom scorer it's a bit unclear.

+1. But maybe with crazy ideas of having pipeline accept a scorer it's
possible.

GaelVaroquaux: > Maybe the pipeline should have a (private?) function to produce the true
> target?
> It think we should solve this as part of the proposal. Otherwise we can't
> really use the pipeline for GridSearchCV or cross_val_score

Pipeline get's scorer as constructor parameter?

mblondel:
It's not clear to me yet why the contract "n_samples doesn't change" is so useful. If we have a way to detect incorrect pipeline usage and do informative error reporting, maybe this contract is not so useful.

Assuming it's ok to change n_samples, how about the following method names:
* fit(X, y=None)
* transform_X(X)
* transform_y(y)
* transform_Xy(X, y)
* transform(X) (alias for transform_X)

(I don't fully comprehend the issues at stake here so sorry if my questions / comments are irrelevant)

mblondel:
Another pattern not discussed so far is:
- scale y
- fit
- predict
- scale back y_pred

The pipeline needs a way to know that the step concerns y and the last step is currently not covered in pipelines. However, I am not sure whether there are other use cases beyond simple y scaling.

mblondel:
Bearing in mind that objects could be used outside of pipelines, I think `fit_filter` is a better name than `fit_pipe`.

amueller:
@mblondel I agree that "scaling y and scaling back after prediction" is a useful pattern. It is currently not included in this proposal, as a pipeline can not operate on the prediction of a previous step (only the last step can use "predict"). We could pretend the prediction is a new ``X`` but that is ugly.

datnamer:
As an end user my dream would be to just wrap a [Blaze](https://github.com/blaze/blaze) data source and have it work in some source of out of core pipeline.

podshumok:
Is it dead?

amueller:
no, it's unfinished ;)

kingjr:
We're facing some difficulties in MNE-Python related to the present proposal and @GaelVaroquaux suggested that it may be useful to describe our usecases so as to motivate your decision. I hope this won't be too irrelevant.

### 1. Initial steps needs all samples, subsequent need a subset of samples

We often need to apply an estimator on a sliding window over `X` of shape `(n_samples, n_features)` where the samples are consecutive "time samples" and are locally correlated with one another.  We're thus constructing a feature space by using adjacent samples. Consequently we're trying to fit a single `coef` vector of shape `(n_time_samples * n_features)` which predict a `y` vector (this is analogous to auto-regressive models). This is a non-conventional design in itself, but the relevant part wrt to the present issue comes next.

Specifically, to construct such rolling time window, we thus needs all samples. However, in the following steps, many samples happen to be detrimental to the fitting (e.g. typically, many samples have all their features = 0). We thus need to exclude these samples from the subsequent steps in the pipeline.

### 2. Semi supervision

It is not entirely clear to me how/whether the current proposal suggests to inform each step of the pipeline about whether each values of `y` was known, or has been modified by a previous step. 

Specifically, one may need to do:
```
pipe = make_pipeline_xy(
      EstimatorFittedOnKnownYOnly(transform_values=np.nan),  # modified y==np.nan by prediction
      EstimatorFittedOnModifiedY(weight_y_known=.80),  # gives higher weights for known than modified y
)
```

### 3. Scoring

I'm unclear about what will be the 'scoring' consequences of your each alternative of the proposal.

So far, I've systematically written meta-estimators to handle these cases, but it's true that they tend to become a bit nested, and difficult to handle.

glemaitre:
During the development of the [imbalanced-learn](https://github.com/scikit-learn-contrib/imbalanced-learn), we actually confront the issue since we need to modify both `X` and `y`.

After discussing briefly with @GaelVaroquaux, we found the right discussion thread.
Therefore, we thought that it could be a good idea to present succinctly the choice of API that we did.
Subsequently, our aim is threefold: (i) we would like to be sure that our API will follow the choice of the scikit-learn community when such transformer will get in the toolbox; (ii) we would like to transfer all implemented algorithms which meet the inclusion criteria of scitkit-learn; and (iii) we would like to continue implementing "cutting-edge" methods which would be complementary to scikit-learn.

## 1. Base class

### 1.1 `SamplerMixin`

We decided to create a new `Mixin` estimator called [`SamplerMixin`](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/imblearn/base.py#L18) since that all the method aimed at resampling `X` and `y`. The estimator is made of 3 methods:

- [`fit`](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/imblearn/base.py#L48): we use this method for checking and computing statistics which are needed at `sample` time. In fact, all these processes could be moved in `sample` for the current algorithm, which come back to the discussion at 2.2.2.1.
- [`sample`](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/imblearn/base.py#L115): we use this method to perform the resampling.
- [`fit_sample`](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/imblearn/base.py#L154): `fit` followed by `sample`.

### 1.2 `BaseBinarySampler` vs. `BaseMulticlassSampler`

Since that some of the algorithms are not designed to work with multiple classes, we created 2 base classes inheriting from `SamplerMixin` to check for the target type and potentially raise warnings. This is permissive for the moment to be compatible with the scikit-learn base estimator.

## 2. Pipeline object

We modified the scikit-learn [`Pipeline`](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/imblearn/pipeline.py) to take into account any sampler in a pipeline.
The sampling is only performed at `fit` time and not at `predict` time. It appears that this is what one would actually want to do while correcting balancing problem; change the balancing at training but test on an imbalanced dataset.
Since `TransformerMixin` and `SamplerMixin` are distincts, we didn't got into trouble with the point 2.2.2.2-2 but at the cost of a new base class.

I hope that I don't forget anything. @chkoar @dvro correct me if this is the case.

dvro:
@glemaitre I'd also add that [imbalanced-learn](https://github.com/scikit-learn-contrib/imbalanced-learn) implementation enables us to work with different approaches. Given a set of samples S = (X, y):

1. UndersamplingPrototype Selection: returns a new set of samples S', S' in S.
2. Undersampling/Prototype Generation: returns a new set of samples S', |S'| < |S|.
3. Oversampling/Selection: returns a new set of samples S', |S'| > |S|.
4. Oversampling/Generation:  returns a new set of samples S', S in S'.
5. Relabeling: returns the same a relabeled set of samples S'.

As you can see, some of these transformations could be implemented by returning a list of index, so that, `X, y = X[idx,:], y[idx]`. However, other transformations actually require returning a new X and y (Prototype Generation and Relabeling). Therefore, (the way I see it) a `fit_filter` is would not work in this second case.

jnothman:
I think we have a few distinct cases that should be handled separately (if at all).

# Data reduction

Resampling and outlier removal at fit time can be done with CV splitters (or by transforming `sample_weight`, which is another question...). They're not a particularly interesting case. Other data reduction, e.g. via BIRCH, is interesting, and yet it makes sense for all these cases to share an interface and be supported by `Pipeline`.

I think for all of these cases imbalanced-learn has got it more-or-less right. Sampler objects should not do anything at predict/transform time, only at `fit` time. From the perspective of `Pipeline`, only `fit_sample` matters.  @glemaitre, do you have good use-cases for having a separate `sample` method? I can imagine its use in out-of-core learning where we want to maintain outlier statistics for a first batch, but again pipeline does not support this case.

Until we find a compelling use-case, I think `fit_sample` and `[fit_]transform` on the same object should be disallowed by `Pipeline`. I am inclined to implement the imbalanced-learn API.

# Target transformation

Transforming targets, as in "scaling y and scaling back after prediction" [mentioned above](https://github.com/scikit-learn/enhancement_proposals/pull/2#issuecomment-150923159), seems a better fit for a meta-estimator, as it:
* bookends a particular process (applies before and after)
* is just as likely to be applied around a single transformer or predictor as around a pipeline subsequence
* does not seem appropriate to append the reverse transformation to a pipeline as this changes the special semantics of the last step of a pipeline.

I think we should have a ?`TransformedTargetPredictor`:

```
class TransformedTargetPredictor(BaseEstimator):
    def __init__(self, estimator, func, inverse_func=None):
        ...
    def fit(self, X, y=None, **kwargs):
        self.estimator_ = clone(self.estimator)
        self.estimator_.fit(X, func(y))
        return self
    def predict(self, X):
        y = self.estimator_.predict(X)
        if self.inverse_func is None:
            return y
        return self.inverse_func(y)
```

This would at least establish a best practice for such things, allowing people to build similar transformations (for what use case, I don't know) dependent on `X, y`.

# Data conversion

This is the challenging case, as meta-estimators do not feel appropriate. However, it seems that we're breaking the intuition of our pipeline/scoring design if we allow this kind of transformation within a `Pipeline`. The contract that the output of a `Pipeline`'s `predict` should look like the `y` that is passed into its `fit` seems an essential one to me. I'm therefore not convinced that it deserves sharing the same API as resampling. Hmm.

glemaitre:
>  @glemaitre, do you have good use-cases for having a separate `sample` method? 

I would say no. A `fit_sample` seems more appropriate as you suggested.

jnothman:
Btw, an implementation and example of using a meta-CV splitter for resampling is at https://gist.github.com/jnothman/6bddbbcca71bdf9fd37e8495d70b42e8. Again, the main problem is its not supporting other training modification such as data reduction or insertion of label noise.

jnothman:
FWIW, https://github.com/dvro/scikit-protopy/blob/master/protopy has an equivalent of "fit_[re]sample" called "reduce_data"

chkoar:
@jnothman yes. @dvro is the author of the `scikit-protopy` and he has imnplemented some of the `imbalanced-learn` algorithms according to the `fit_sample` interface.

jnothman:
One arguable use case not considered here is (parameter search in) semi-supervised learning. Assume we want to select the LDA model over a large collection of texts which maximises text classification performance with a small amount of labelled data. In this case, the transformers need the unlabelled data at fit time, but the classifier does not, nor does predict time.

A potential, if perhaps unparadigmatic, solution may combine the following:
```python
class DiscardUnlabelled(BaseEstimator):
    def fit_resample(self, X, y):
        return X[y != -1], y[y != -1]

class SemiSupervisedSplit:
    def __init__(self, base_cv):
        self.base_cv = base_cv
    def split(self, X, y):
        has_label = y != -1
        has_label_idx = np.flatnonzero(has_label)
        no_label_idx = np.flatnonzero(~has_label)
        for train_idx, test_idx in self.base_cv.split(X[has_label_idx], y[has_label_idx]):
            yield np.concatenate(no_label_idx, has_label_idx[train_idx]), has_label_idx[test_idx]
```

amueller:
I feel more and more that the most pragmatic solution might be by now to
not introduce a new method, but instead allow X to be DataFrame that also
contains the labels. That doesn't solve everything but it makes for a much
more natural workflow for most people.

Sent from phone. Please excuse spelling and brevity.

On Feb 27, 2017 8:15 PM, "Joel Nothman" <notifications@github.com> wrote:

One arguable use case not considered here is (parameter search in)
semi-supervised learning. Assume we want to select the LDA model over a
large collection of texts which maximises text classification performance
with a small amount of labelled data. In this case, the transformers need
the unlabelled data at fit time, but the classifier does not, nor does
predict time.

A potential, if perhaps unparadigmatic, solution may combine the following:

class DiscardUnlabelled(BaseEstimator):
    def fit_resample(self, X, y):
        return X[y != -1], y[y != -1]
class SemiSupervisedSplit:
    def __init__(self, base_cv):
        self.base_cv = base_cv
    def split(self, X, y):
        has_label = y != -1
        has_label_idx = np.flatnonzero(has_label)
        no_label_idx = np.flatnonzero(~has_label)
        for train_idx, test_idx in
self.base_cv.split(X[has_label_idx], y[has_label_idx]):
            yield np.concatenate(no_label_idx,
has_label_idx[train_idx]), has_label_idx[test_idx]

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/scikit-learn/enhancement_proposals/pull/2#issuecomment-282910463>,
or mute the thread

GaelVaroquaux:
> I feel more and more that the most pragmatic solution might be by now to
> not introduce a new method, but instead allow X to be DataFrame that also
> contains the labels. That doesn't solve everything but it makes for a much
> more natural workflow for most people.

I see a few limitations:

* What if X is sparse (or anything others than columns, eg free text)?

* What if y is more complex than a vector? (multi label, for instance)

* I worry a lot about data leakage in cross-validation. Scikit-learn
  is quite good at preventing that (and people keep complaining that they
  cannot do their weird cross-validation thing that has implicit leakage
  :) ).

amueller:
Sparse X and text can be handled by DataFrames, tough I haven't recently
checked on how good the sparse support is. Multilabel y could be handled by
a hierarchical index.
I agree that separating X and y less than our current API might have
downsides. More flexibility allows people to screw up more.
Can you give an example of the cross-validation?
The .split method had access to y so you should be able to do anything.

GaelVaroquaux:
> Sparse X and text can be handled by DataFrames, tough I haven't recently checked on how good the sparse support is.

Not great, AFAIK, but I'd love to be proven wrong.

> Multilabel y could be handled by a hierarchical index.

It makes things much harder and much more clumsy. Pandas is powerful, but
hard to use.

> Can you give an example of the cross-validation?

Mixing X and y in a transformer.

amueller:
Text was never a problem iirc. Sparse data requires a copy and looks a bit half-baked: http://pandas.pydata.org/pandas-docs/stable/sparse.html

I don't think hierarchical indices are clumsy, though they require some understanding of pandas.

GaelVaroquaux:
> I don't think hierarchical indices are clumsy, though they require some understanding of pandas.

My students keep making errors that are very very hard to debug (I do
too). They are handy, but lead to woodoo-style code. I don't think that
it's a good basics for an API for scikit-learn.

jnothman:
I don't see how this fixes things, for many cases. One major feature of
pipeline is that it trains the final predictor to output predictions
comparable to the evaluation ground truth. How would this be ensured if a
dataframe is used?

PR #17, opened 2019-04-05
Slep007 - feature names, their generation and the API

adrinjalali:
Since it seems the discussion in https://github.com/scikit-learn/scikit-learn/pull/13307 requires a SLEP to decide on the style of feature names, and we talked about lowering the bar for a submitting a slep, I took the liberty to open one for us to discuss the styles and hopefully soon come to a consensus/conclusion.

### DIFF
--- a/slep007/proposal.rst
+++ b/slep007/proposal.rst
@@ -8,51 +8,49 @@ Feature Names
 proposes a solution to support and propagate feature names through a pipeline.
 
 However, the generated feature names can become long. This is to confirm how
-we want those generated feature names to behave, and this is the proposal:
+we want those generated feature names to behave, and this is the proposal::
 
 
-```python
-from sklearn.pipeline import make_pipeline
-from sklearn.preprocessing import StandardScaler
-from sklearn.decomposition import PCA
-from sklearn.compose import make_column_transformer
-from sklearn.datasets import load_iris
-from sklearn.linear_model import LogisticRegression
-from sklearn.feature_selection import SelectKBest
-import pandas as pd
+    from sklearn.pipeline import make_pipeline
+    from sklearn.preprocessing import StandardScaler
+    from sklearn.decomposition import PCA
+    from sklearn.compose import make_column_transformer
+    from sklearn.datasets import load_iris
+    from sklearn.linear_model import LogisticRegression
+    from sklearn.feature_selection import SelectKBest
+    import pandas as pd
 
-iris = load_iris()
-X = pd.DataFrame(iris.data, columns=iris.feature_names)
-pipe = make_pipeline(StandardScaler(), PCA(), SelectKBest(k=2),
-                     LogisticRegression())
-pipe.fit(X, iris.target)
-pipe[-1].input_features_
-```
-> array(['pca0', 'pca1'], dtype='<U4')
+    iris = load_iris()
+    X = pd.DataFrame(iris.data, columns=iris.feature_names)
+    pipe = make_pipeline(StandardScaler(), PCA(), SelectKBest(k=2),
+                         LogisticRegression())
+    pipe.fit(X, iris.target)
+    pipe[-1].input_features_
+    > array(['pca0', 'pca1'], dtype='<U4')
 
-```python
-# I have to duplicate StandardScaler if I want to be able to
-# use the pandas column names in the column transformer. Yuk!
-# also there's no easy way to pass through the original columns with
-# ColumnTransformer - FunctionTransformer would remove feature names!
-pipe = make_pipeline(make_column_transformer((make_pipeline(StandardScaler(),
-                                                            PCA()), X.columns),
-                                             (StandardScaler(),
-                                             X.columns[:2])),
-                     SelectKBest(k=2), LogisticRegression())
-pipe.fit(X, iris.target)
-pipe[-1].input_features_
-```
-> array(['pipeline__pca0', 'standardscaler__sepal length (cm)'], dtype='<U33')
-```python
-pipe = make_pipeline(make_column_transformer((PCA(), X.columns),
-                                             (StandardScaler(),
-                                             X.columns[:2])),
-                     SelectKBest(k=2), LogisticRegression())
-pipe.fit(X, iris.target)
-pipe[-1].input_features_
-```
-> array(['pca__pca0', 'standardscaler__sepal length (cm)'], dtype='<U33')
+
+    # I have to duplicate StandardScaler if I want to be able to
+    # use the pandas column names in the column transformer. Yuk!
+    # also there's no easy way to pass through the original columns with
+    # ColumnTransformer - FunctionTransformer would remove feature names!
+    pipe = make_pipeline(make_column_transformer((make_pipeline(StandardScaler(),
+                                                                PCA()), X.columns),
+                                                 (StandardScaler(),
+                                                 X.columns[:2])),
+                         SelectKBest(k=2), LogisticRegression())
+    pipe.fit(X, iris.target)
+    pipe[-1].input_features_
+    ```
+    > array(['pipeline__pca0', 'standardscaler__sepal length (cm)'], dtype='<U33')
+
+    pipe = make_pipeline(make_column_transformer((PCA(), X.columns),
+                                                 (StandardScaler(),
+                                                 X.columns[:2])),
+                         SelectKBest(k=2), LogisticRegression())
+    pipe.fit(X, iris.target)
+    pipe[-1].input_features_
+
+    > array(['pca__pca0', 'standardscaler__sepal length (cm)'], dtype='<U33')
 
 Is that what we want? (apart from changing to object dtype lol)
 The first one seems good to me, the others seem a bit long? Not sure how to do
--- a/slep007/proposal.rst
+++ b/slep007/proposal.rst
@@ -1,0 +1,59 @@
+ .. _slep_007:
+
+=============
+Feature Names
+=============
+
+`scikit-learn/#13307 <https://github.com/scikit-learn/scikit-learn/pull/13307`_
+proposes a solution to support and propagate feature names through a pipeline.
+
+However, the generated feature names can become long. This is to confirm how
+we want those generated feature names to behave, and this is the proposal:
+
+
+```python
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+from sklearn.decomposition import PCA
+from sklearn.compose import make_column_transformer
+from sklearn.datasets import load_iris
+from sklearn.linear_model import LogisticRegression
+from sklearn.feature_selection import SelectKBest
+import pandas as pd
+
+iris = load_iris()
+X = pd.DataFrame(iris.data, columns=iris.feature_names)
+pipe = make_pipeline(StandardScaler(), PCA(), SelectKBest(k=2),
+                     LogisticRegression())
+pipe.fit(X, iris.target)
+pipe[-1].input_features_
+```
+> array(['pca0', 'pca1'], dtype='<U4')
+
+```python
+# I have to duplicate StandardScaler if I want to be able to
+# use the pandas column names in the column transformer. Yuk!
+# also there's no easy way to pass through the original columns with
+# ColumnTransformer - FunctionTransformer would remove feature names!
+pipe = make_pipeline(make_column_transformer((make_pipeline(StandardScaler(),
+                                                            PCA()), X.columns),
+                                             (StandardScaler(),
+                                             X.columns[:2])),
+                     SelectKBest(k=2), LogisticRegression())
+pipe.fit(X, iris.target)
+pipe[-1].input_features_
+```
+> array(['pipeline__pca0', 'standardscaler__sepal length (cm)'], dtype='<U33')
+```python
+pipe = make_pipeline(make_column_transformer((PCA(), X.columns),
+                                             (StandardScaler(),
+                                             X.columns[:2])),
+                     SelectKBest(k=2), LogisticRegression())
+pipe.fit(X, iris.target)
+pipe[-1].input_features_
+```
+> array(['pca__pca0', 'standardscaler__sepal length (cm)'], dtype='<U33')
+
+Is that what we want? (apart from changing to object dtype lol)
+The first one seems good to me, the others seem a bit long? Not sure how to do
+better though.
--- a/slep007/proposal.rst
+++ b/slep007/proposal.rst
@@ -71,13 +71,21 @@ Output Feature Names
 --------------------
 
 A fitted estimator exposes the output feature names through the
-``feature_names_out_`` attribute.
+``feature_names_out_`` attribute. Here we discuss more in detail how these
+feature names are generated. Since for most estimators there are multiple ways
+to generate feature names, this SLEP does not intend to define how exactly
+feature names are generated for all of them. It is instead a guideline on how
+they could generally be generated.
 
 One-to-one Transformers
 ***********************
 
 From the perspective of feature names, the simplest transformers are the ones
-which have the same feature names in the output as in the input.
+which have the same feature names in the output as in the input, and the
+transformation done on the data is semi-trivial. The ``StandardScaler`` can be
+one example. However, a transformer can choose to be more verbose and generate
+a more informative feature name, ``scaled(income)`` could be an example, and
+the verbosity is controlled by a parameter to the estimator.
 
 - Input provides feature names: ``feature_names_in_`` and
   ``feature_names_out_`` are the same.
@@ -89,8 +97,9 @@ Feature Selector Transformers
 
 The output feature names are the ones selected from the input, and if no
 feature names are provided, ``x0`` to ``xn`` are assumed to be their names. For
-example, if the transformer selects the first and the third features, and no
-names are provided, the ``feature_names_out_`` will be ``[x0, x2]``.
+example, if a ``SelectKBest`` transformer selects the first and the third
+features, and no names are provided, the ``feature_names_out_`` will be ``[x0,
+x2]``.
 
 Feature Generating Transformers
 *******************************
@@ -123,7 +132,9 @@ the same as the ``feature_names_out_`` of the last step, and ``None`` if the
 last step is not a transformer.
 
 ``ColumnTransformer`` by default adds a prefix to the output feature names,
-indicating the name of the step applied on them.
+indicating the name of the step applied on them. If a column is in the output
+as a part of ``passthrough``, it won't be prefixed since no operation has been
+applied on it.
 
 This is the default behavior, and it can be tuned by constructor parameters if
 the meta estimator allows it. For instance, a ``prefix=False`` may indicate
@@ -141,4 +152,6 @@ developers to implement the API.
 Notes
 -----
 
-This SLEP also applies to resamplers the same way as transformers.
+This SLEP also applies to `resamplers
+<https://github.com/scikit-learn/enhancement_proposals/pull/15>`_ the same way
+as transformers.
--- a/slep007/proposal.rst
+++ b/slep007/proposal.rst
@@ -12,9 +12,9 @@ Feature names, their generation and the API
 Abstract
 ########
 
-This SLEP proposes the introduction of the `feature_names_in_` attribute for
-all estimators, and the `feature_names_out_` attribute for all transformers. We
-here discuss the generation of such attributes and their propagation through
+This SLEP proposes the introduction of the ``feature_names_in_`` attribute for
+all estimators, and the ``feature_names_out_`` attribute for all transformers.
+We here discuss the generation of such attributes and their propagation through
 pipelines.
 
 Motivation
@@ -117,11 +117,9 @@ attribute for all estimators, and a ``feature_names_out_`` attribute for any
 estimator with a ``transform`` method, *i.e.* they expose the generated feature
 names via the ``feature_names_out_`` attribute.
 
-.. note:
-
-    This SLEP also applies to `resamplers
-    <https://github.com/scikit-learn/enhancement_proposals/pull/15>`_ the same
-    way as transformers.
+Note that this SLEP also applies to `resamplers
+<https://github.com/scikit-learn/enhancement_proposals/pull/15>`_ the same way
+as transformers.
 
 Input Feature Names
 ###################
@@ -139,16 +137,20 @@ A fitted estimator exposes the output feature names through the
 feature names are generated. Since for most estimators there are multiple ways
 to generate feature names, this SLEP does not intend to define how exactly
 feature names are generated for all of them. It is instead a guideline on how
-they could generally be generated.
+they could generally be generated. Furthermore, that specific behavior of a
+given estimator may be tuned via the ``verbose_feature_names`` parameter, as
+detailed below.
+
+When generating the output feature names, if no input feature names are
+provided, ``x0`` to ``xn`` are assumed to be their names.
 
 Feature Selector Transformers
 *****************************
 
-The output feature names are the ones selected from the input, and if no
-feature names are provided, ``x0`` to ``xn`` are assumed to be their names. For
-example, if a ``SelectKBest`` transformer selects the first and the third
-features, and no names are provided, the ``feature_names_out_`` will be ``[x0,
-x2]``.
+This includes transformers which output a subset of the input features, w/o
+changing them. For example, if a ``SelectKBest`` transformer selects the first
+and the third features, and no names are provided, the ``feature_names_out_``
+will be ``[x0, x2]``.
 
 Feature Generating Transformers
 *******************************
@@ -158,13 +160,13 @@ generate a column based on a single given column. The generated output column
 in this case is a sensible transformation of the input feature name. For
 instance, a ``LogTransformer`` can do ``'age' -> 'log(age)'``, and a
 ``OneHotEncoder`` could do ``'gender' -> 'gender_female', 'gender_fluid',
-...``. An alternative generated feature name for the transformers where each
-output feature corresponds to exactly one input feature is to leave the feature
-name unchanged. Whether or not to modify the feature name, *e.g.* ``log(x0)``
-vs. ``x0`` is controlled via the ``verbose_feature_names`` to the constructor.
-The default value of ``verbose_feature_names`` can be different depending on
-the transformer. For instance, ``StandardScaler`` can have it as ``False``,
-whereas ``LogTransformer`` could have it as ``True`` by default.
+...``. An alternative is to leave the feature names unchanged when each output
+feature corresponds to exactly one input feature. Whether or not to modify the
+feature name, *e.g.* ``log(x0)`` vs. ``x0`` may be controlled via the
+``verbose_feature_names`` to the constructor. The default value of
+``verbose_feature_names`` can be different depending on the transformer. For
+instance, ``StandardScaler`` can have it as ``False``, whereas
+``LogTransformer`` could have it as ``True`` by default.
 
 Transformers where each output feature depends on a fixed number of input
 features may generate descriptive names as well. For instance, a
@@ -226,12 +228,12 @@ names::
     feature_names_out_: [pca0, pca1, pca2]
 
 
-    [model, make, num0, ..., num100] ->
+    [model, make, numeric0, ..., numeric100] ->
         ColumnTransformer(
             [('cat', Pipeline(SimpleImputer(), OneHotEncoder()),
               ['model', 'make']),
              ('num', Pipeline(SimpleImputer(), PCA(n_components=3)),
-              ['num0', ..., 'num100'])]
+              ['numeric0', ..., 'numeric100'])]
         )
     feature_names_out_: ['cat_model_100', 'cat_model_200', ...,
                          'cat_make_ABC', 'cat_make_XYZ', ...,
@@ -240,10 +242,10 @@ names::
 However, the following examples produce a somewhat redundant feature names,
 and hence the relevance of ``verbose_feature_names=False``::
 
-    [model, make, num0, ..., num100] ->
+    [model, make, numeric0, ..., numeric100] ->
         ColumnTransformer([
             ('ohe', OneHotEncoder(), ['model', 'make']),
-            ('pca', PCA(n_components=3), ['num0', ..., 'num100'])
+            ('pca', PCA(n_components=3), ['numeric0', ..., 'numeric100'])
         ])
     feature_names_out_: ['ohe_model_100', 'ohe_model_200', ...,
                          'ohe_make_ABC', 'ohe_make_XYZ', ...,
@@ -251,10 +253,10 @@ and hence the relevance of ``verbose_feature_names=False``::
 
 If desired, the user can remove the prefixes::
 
-    [model, make, num0, ..., num100] ->
+    [model, make, numeric0, ..., numeric100] ->
         make_column_transformer(
             (OneHotEncoder(), ['model', 'make']),
-            (PCA(n_components=3), ['num0', ..., 'num100']),
+            (PCA(n_components=3), ['numeric0', ..., 'numeric100']),
             verbose_feature_names=False
         )
     feature_names_out_: ['model_100', 'model_200', ...,
@@ -266,5 +268,5 @@ Backward Compatibility
 
 All estimators should implement the ``feature_names_in_`` and
 ``feature_names_out_`` API. This is checked in ``check_estimator``, and the
-transition is done with a ``FutureWarning`` to give time to third party
-developers to implement the API.
+transition is done with a ``FutureWarning`` for at least two versions to give
+time to third party developers to implement the API.
--- a/slep007/proposal.rst
+++ b/slep007/proposal.rst
@@ -6,11 +6,114 @@ Feature Names
 
 ``scikit-learn`` has been making it easier to build complex workflows with the
 ``ColumnTransformer`` and it has been seeing widespread adoption. However,
-using it results in very opaque models, even more so than before. There is a
-`usage example
-<https://scikit-learn.org/dev/auto_examples/compose/plot_column_transformer_mixed_types.html>`_
-in the gallery that applies a classifier to the titanic data set, which is a
-very simple standard usecase.
+using it results in pipelines where it's not clear what the input features to
+the final predictor are, even more so than before. The following example
+illustrates that very well::
+
+
+    """
+    ===================================
+    Column Transformer with Mixed Types
+    ===================================
+
+    This example illustrates how to apply different preprocessing and
+    feature extraction pipelines to different subsets of features,
+    using :class:`sklearn.compose.ColumnTransformer`.
+    This is particularly handy for the case of datasets that contain
+    heterogeneous data types, since we may want to scale the
+    numeric features and one-hot encode the categorical ones.
+
+    In this example, the numeric data is standard-scaled after
+    mean-imputation, while the categorical data is one-hot
+    encoded after imputing missing values with a new category
+    (``'missing'``).
+
+    Finally, the preprocessing pipeline is integrated in a
+    full prediction pipeline using :class:`sklearn.pipeline.Pipeline`,
+    together with a simple classification model.
+    """
+
+    # Author: Pedro Morales <part.morales@gmail.com>
+    #
+    # License: BSD 3 clause
+
+    import numpy as np
+
+    from sklearn.compose import ColumnTransformer
+    from sklearn.datasets import fetch_openml
+    from sklearn.pipeline import Pipeline
+    from sklearn.impute import SimpleImputer
+    from sklearn.preprocessing import StandardScaler, OneHotEncoder
+    from sklearn.linear_model import LogisticRegression
+    from sklearn.model_selection import train_test_split, GridSearchCV
+
+    np.random.seed(0)
+
+    # Load data from https://www.openml.org/d/40945
+    X, y = fetch_openml("titanic", version=1, as_frame=True, return_X_y=True)
+
+    # Alternatively X and y can be obtained directly from the frame attribute:
+    # X = titanic.frame.drop('survived', axis=1)
+    # y = titanic.frame['survived']
+
+    # We will train our classifier with the following features:
+    # Numeric Features:
+    # - age: float.
+    # - fare: float.
+    # Categorical Features:
+    # - embarked: categories encoded as strings {'C', 'S', 'Q'}.
+    # - sex: categories encoded as strings {'female', 'male'}.
+    # - pclass: ordinal integers {1, 2, 3}.
+
+    # We create the preprocessing pipelines for both numeric and categorical data.
+    numeric_features = ['age', 'fare']
+    numeric_transformer = Pipeline(steps=[
+        ('imputer', SimpleImputer(strategy='median')),
+        ('scaler', StandardScaler())])
+
+    categorical_features = ['embarked', 'sex', 'pclass']
+    categorical_transformer = Pipeline(steps=[
+        ('imputer', SimpleImputer(strategy='constant', fill_value='missing')),
+        ('onehot', OneHotEncoder(handle_unknown='ignore'))])
+
+    preprocessor = ColumnTransformer(
+        transformers=[
+            ('num', numeric_transformer, numeric_features),
+            ('cat', categorical_transformer, categorical_features)])
+
+    # Append classifier to preprocessing pipeline.
+    # Now we have a full prediction pipeline.
+    clf = Pipeline(steps=[('preprocessor', preprocessor),
+                          ('classifier', LogisticRegression())])
+
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2)
+
+    clf.fit(X_train, y_train)
+    print("model score: %.3f" % clf.score(X_test, y_test))
+
+
+    ###############################################################################
+    # Using the prediction pipeline in a grid search
+    ###############################################################################
+    # Grid search can also be performed on the different preprocessing steps
+    # defined in the ``ColumnTransformer`` object, together with the classifier's
+    # hyperparameters as part of the ``Pipeline``.
+    # We will search for both the imputer strategy of the numeric preprocessing
+    # and the regularization parameter of the logistic regression using
+    # :class:`sklearn.model_selection.GridSearchCV`.
+
+
+    param_grid = {
+        'preprocessor__num__imputer__strategy': ['mean', 'median'],
+        'classifier__C': [0.1, 1.0, 10, 100],
+    }
+
+    grid_search = GridSearchCV(clf, param_grid, cv=10)
+    grid_search.fit(X_train, y_train)
+
+    print(("best logistic regression from grid search: %.3f"
+           % grid_search.score(X_test, y_test)))
+
 
 However, it's impossible to interpret or even sanity-check the
 ``LogisticRegression`` instance that's produced in the example, because the
@@ -35,20 +138,21 @@ abovementioned example ``clf[-1].feature_names_in_`` and
      'cat__pclass_3']
 
 Ideally the generated feature names describe how a feature is generated at each
-stage. For instance, ``cat__sex_female`` shows that the feature has been
-through a categorical preprocessing pipeline, was originally the column
-``sex``, and has been one hot encoded and is one if it was originally
+stage of a pipeline. For instance, ``cat__sex_female`` shows that the feature
+has been through a categorical preprocessing pipeline, was originally the
+column ``sex``, and has been one hot encoded and is one if it was originally
 ``female``. However, this is not always possible or desirable especially when a
-generated column is based on many columns, for example in ``PCA``. As a rule of
-thumb, the following types of transformers may generate feature names which
-corresponds to the original features:
+generated column is based on many columns, since the generated feature names
+will be too long, for example in ``PCA``. As a rule of thumb, the following
+types of transformers may generate feature names which corresponds to the
+original features:
 
 - Leave columns unchanged, *e.g.* ``StandardScaler``
 - Select a subset of columns, *e.g.* ``SelectKBest``
 - create new columns where each column depends on at most one input column,
   *e.g* ``OneHotEncoder``
 - Algorithms that create combinations of a fixed number of features, *e.g.*
-  ``PolynomialFeatures`` on a fixed subset of features, as opposed to all of
+  ``PolynomialFeatures``, as opposed to all of
   them where there are many. Note that verbosity considerations and
   ``prefix_feature_names`` as explained later can apply here.
 
@@ -58,8 +162,9 @@ propagated.
 Scope
 -----
 
-All estimators implement a ``feature_names_in_`` API, and any estimator with
-a ``transform`` method implements a ``feature_names_out_`` API.
+All estimators implement a ``feature_names_in_`` API, and any estimator with a
+``transform`` method implements a ``feature_names_out_`` API, _i.e._ they
+expose the generated feature names via the ``feature_names_out_`` attribute.
 
 Input Feature Names
 -------------------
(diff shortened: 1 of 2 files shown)


amueller:
My biggest concern right now is what the feature names should actually be, and I think my doubts are mostly around creating new feature names within pipelines and column transformer.

amueller:
I don't like the ``pca__pca0`` name because it's redundant and seems like it would be  a common occurrence. But maybe I'm too worried about it?

Maybe explaining better the mechanism for feature names would make it clearer?
I had posted some discussion in https://github.com/scikit-learn/scikit-learn/pull/12627

There I talked about three cases:

a) The transformer does a non-trivial column transformation, like OneHotEncoder or feature selection
b) The transformer does "nothing" to the columns, like StandardScaler.
c) The transformer does a "too complex" operation to the columns, like PCA.

I guess a slight confusion I have right now is because there is another case, which is the ``ColumnTransformer``, which also creates new feature names by first prepending the transformer name, and then passing through the column names - that's different than pipelines which don't prepend the estimator name.

My question right now is: do we always prepend the estimator name in ColumnTransformer, even if it generates things like ``pca__pca0`` or ``standardscaler__sepallength``.

adrinjalali:
regarding the redundancy of the feature names, we could
- only add the prefix if there are duplicate names (which can also be problematic in some edge cases)
- allow the user to set a function somewhere which would do the transformation based on the users' preferences.

amueller:
Not sure I really get the 2nd.
The 1st is basically an if in the column transformer that checks if a column is used multiple times. And then what?

GaelVaroquaux:
I've been playing with dask and pandas on very large datasets with many inconsistencies. The experience has really reminded me that code that does not have very consistent behavior leads to nasty bugs.

slep007/proposal.rst
```diff
@@ -0,0 +1,57 @@
...
+    X = pd.DataFrame(iris.data, columns=iris.feature_names)
+    pipe = make_pipeline(StandardScaler(), PCA(), SelectKBest(k=2),
+                         LogisticRegression())
+    pipe.fit(X, iris.target)
+    pipe[-1].input_features_
+    > array(['pca0', 'pca1'], dtype='<U4')
+
+
+    # I have to duplicate StandardScaler if I want to be able to
+    # use the pandas column names in the column transformer. Yuk!
```
GaelVaroquaux: Couldn't you put the standard scaler in front of the column transformer here?

adrinjalali: I think the idea is that "the user could do this", just to showcase what happens if they do. Not that it's the best way to do it.

amueller: No, you can not, as the comment suggests: this proposal is about creating feature names, not doing "pandas in, pandas out". StandardScaler will therefore create a numpy array as output, so ColumnTransformer can not use column names. A work-around would be to use column indices in the column transformer, or a boolean mask (explicitly using knowledge that StandardScaler will preserve columns).

amueller:
@GaelVaroquaux I'm not sure I understand what your comment about inconsistent behavior is about. Was that re @adrinjalali's suggestion of sometimes not appending the name?

amueller:
@adrinjalali The second one doesn't really make sense the way you wrote it, I think. If you want to replicate the prefix behavior, ``fname_transformer`` also needs to get the transformer name at least. I guess that's what you meant?
Are there other functions that would be sensible here apart from ``lambda t_name, f_name: t_name + f_name`` and ``lambda t_name, f_name: f_name``?

I don't like 1) because it means that a feature name of an existing feature can change when I add additional features. That's quite counter-intuitive.
Though I guess you could argue the same is true for ``make_column_transformer`` names?

GaelVaroquaux:
> @GaelVaroquaux I'm not sure I understand what your comment about inconsistent behavior is about. Was that re @adrinjalali's suggestion of sometimes not appending the name?

Yes.

GaelVaroquaux:
> No, you can not, as the comment suggests: this proposal is about creating feature names, not doing "pandas in, pandas out". StandardScaler will therefore create a numpy array as output, so ColumnTransformer can not use column names.

Point taken. Thank you.

amueller:
I'm a bit annoyed at the duplication of the standard scaler, but I feel it will be hard to make this work better without blowing up the scope of the proposal, and it highlights a shortcoming of the current design; the shortcoming is unrelated to the feature name proposal, though, it's "only" related to dealing with pandas data frames.

adrinjalali:
> Are there other functions that would be sensible here apart from lambda t_name, f_name: t_name + f_name and lambda t_name, f_name: f_name?

Those are the most reasonable options I can think of.

> There's no discussion of what the vectorizers do with their input feature names, if anything. Is that even allowed?

Should we then have a parameter like `output_feature_name_policy` or something, which accepts a few values and based on that we generate the output feature names? It would be consistent throughout the package, so @GaelVaroquaux 's concern would be address I guess, and I suppose you would also fix your concerns with the verbose feature names @amueller ?  Of course by default they're pretty verbose, but by passing an appropriate value to the parameter to some of the transformers/pipelines, the generated feature names would make sense?

> @adrinjalali The second one doesn't really make sense the way you wrote it, I think. If you want to replicate the prefix behavior, fname_transformer also needs to get the transformer name at least. I guess that's what you meant?

Yes, exactly, we can even get it from the `self` if we add a `Mixin`? Shouldn't be too much chage.

amueller:
> Should we then have a parameter like output_feature_name_policy or something, which accepts a few values and based on that we generate the output feature names? 

parameter to what? each estimator? just the ColumnTransformer? global configuration (errrr probably not if column names will be used in actual behavior).

adrinjalali:
> parameter to what? each estimator? just the ColumnTransformer? global configuration (errrr probably not if column names will be used in actual behavior).

A parameter used by `feature_names_out()`. I was thinking of having it in each estimator, it'd be easier for the user to tune the parameters. Otherwise if we have it only for Pipeline and ColumnTransformer, then we'd need to provide a way for the user to specify may be a `{estimator: policy}` dict, which doesn't seem easier from the user side to me.

What we can do is to expose a function `set_ofnames_policy()` in the `Mixin`, so that it can easily be discovered by the user. Then we can start by adding the param to estimators' `__init__`, starting by the ones which would need it more frequently. But it doesn't have to be in all of them to begin with.

amueller:
ok so you mean adding a parameter to the constructor.

I don't understand your argument about having to pass a dict. It's a constructor argument for a single estimator, right? So I know exactly which estimator I'm applying it to.

adrinjalali:
> I don't understand your argument about having to pass a dict. It's a constructor argument for a single estimator, right? So I know exactly which estimator I'm applying it to.

If we accept the extra parameter to the constructor, then we don't need the dict. The dict was the alternative: if we reject the proposal to have the constructor argument, and want to have the policy argument only available to a few objects such as ColumnTransformer, Pipeline, etc, then we need the dict for a more granular control over the policies of each underlying object. To be clear, this is not what I'm proposing and I personally dislike it.

thomasjpfan:
A random thought: with this approach the following will have the same feature names

### SelectKBest before PCA

Three of the original features were passed into the PCA.
```py
pipe = make_pipeline(StandardScaler(), SelectKBest(k=3), PCA(n_components=2),
	                 LogisticRegression())
pipe.fit(X, iris.target)
pipe[-1].input_features_
# array(['pca0', 'pca1'], dtype='<U4')
```

### SelectKBest after PCA

All the original features are passed into PCA, and two of the pca features are selected.
```py
pipe = make_pipeline(StandardScaler(), PCA(), SelectKBest(k=2),
	                 LogisticRegression())
pipe.fit(X, iris.target)
pipe[-1].input_features_
# array(['pca0', 'pca1'], dtype='<U4')
```

amueller:
I don't think that they have the same feature names is an issue. I think the issue is that information is lost in the first case.

adrinjalali:
> A random thought: with this approach the following will have the same feature names

I think that's fine, the combination of feature names with the new pipeline visualizer should be enough for people to follow what happens inside.

adrinjalali:
> @adrinjalali but in the first case there's no way for you to know what the selection did.

`pipe[1].feature_names_out_` will be the original features selected by that transformer. Is that not enough?

amueller:
@adrinjalali That depends a lot on what we consider enough, I think. It means that if the user plots the feature names in the "standard" way by looking at what goes into LogisticRegression, they'll miss part of the story.
But to provide the full information we would need to use much more elaborate/verbose feature names (like ``pca1(x_1, x_2)``). If there's many features this will quickly get out of hand, though...

amueller:
@thomasjpfan you had a PR where you provided full input-output correspondences, right (or was it a comment somewhere)? That would allow tracking "everything" but might become very verbose.

thomasjpfan:
I have a PR for `ColumnTransformer` https://github.com/scikit-learn/scikit-learn/pull/11639, it was needed to make `inverse_transform` work.

slep007/proposal.rst
```diff
@@ -0,0 +1,84 @@
...
+data during fit. It is taken either from a ``pandas`` data frame or our own
+data structure (something on top of a ``numpy`` array, ``NamedArray`` for
+instance). The attribute will be ``None`` if the given data does not include
+any feature names.
+
+Output Feature Names
+--------------------
+
+A fitted estimator exposes the output feature names through the
+``feature_names_out_`` attribute. At the same time, the data structure returned
```
amueller: I thought this wasn't about how they are propagated ;) Do you want to remove the second sentence?

slep007/proposal.rst
```diff
@@ -0,0 +1,84 @@
...
+by a ``transform`` method also includes the generated feature names.
+
+Same Features Transformers
```
amueller: I called them One-to-one which I think is more clear.

slep007/proposal.rst
```diff
@@ -0,0 +1,84 @@
...
+
+From the perspective of feature names, the simplest transformers are the ones
+which have the same feature names in the output as in the input.
+
+- Input provides feature names: ``feature_names_in_`` and
+  ``feature_names_out_`` are the same.
+- Input provides no feature names: ``feature_names_out_`` will be ``x0`` to
+  ``xn``, where ``n`` is the number of features.
+
+Feature Selector Transformers
```
amueller: Do you include the imputers in this?

slep007/proposal.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Meta estimators can choose to prefix the output feature names or not.
+
+By default, ``Pipeline`` adds no prefix, *i.e* its ``feature_names_out_`` is
+the same as the ``feature_names_out_`` of the last step, and ``None`` if the
+last step is not a transformer.
+
+``ColumnTransformer`` by default adds a prefix to the output feature names,
+indicating the name of the step applied on them.
+
```
amueller: Where is MissingIndicator and Polynomial Features and OneHotEncoder? Right now these would be in Feature Generating Transformers and their names would be useless.
I feel like you left out all the interesting ones.

I think breaking it down is nice to be concrete but I would also try to write down a general concept / abstract rule in the beginning.

slep007/proposal.rst
```diff
@@ -0,0 +1,84 @@
...
+All estimators should implement the ``feature_names_in_`` and
+``feature_names_out_`` API. This is checked in ``check_estimator``, and the
+transition is done with a ``FutureWarning`` to give time to third party
+developers to implement the API.
+
+Notes
+-----
+
+This SLEP also applies to resamplers the same way as transformers.
```
amueller: That slep isn't accepted yet, right?

slep007/proposal.rst
```diff
@@ -0,0 +1,84 @@
...
+Meta estimators can choose to prefix the output feature names or not.
+
+By default, ``Pipeline`` adds no prefix, *i.e* its ``feature_names_out_`` is
+the same as the ``feature_names_out_`` of the last step, and ``None`` if the
+last step is not a transformer.
+
+``ColumnTransformer`` by default adds a prefix to the output feature names,
+indicating the name of the step applied on them.
+
+
```
amueller: This is definitely missing examples!

adrinjalali:
I rewrote this SLEP to focus on how the names are generated. I have kept them simple, i.e. the user will not see a very verbose feature name at the end, since I don't think it's feasible to have them as soon as there's a PCA in the pipeline. At least the datasets I have dealt with, always have too many features for that to work.

slep007/proposal.rst
adrinjalali: I think imputers are one-to-one transformers, aren't they?

slep007/proposal.rst
adrinjalali: correct, AFAIK

slep007/proposal.rst
amueller: SimpleImputer drops all-nan columns. You can check my implementations of feature names to see all the special cases I implemented ;)
The larger point is that it's easy to overlook estimators here and so a general story would be good.

slep007/proposal.rst
amueller: I don't have a strong opinion but I find it slightly confusing to refer to things that don't exist yet and might never exist.

amueller:
Can you maybe reread https://github.com/scikit-learn/enhancement_proposals/pull/17#issuecomment-480413603

The SLEP should answer that question and/or address that concern. If we decide to have ``pca_pca0`` and ``standardscaler_sepallength`` as feature names that's ok, but it's something that will be produced by the current proposal and that's not mentioned at all.

slep007/proposal.rst
```diff
@@ -0,0 +1,86 @@
...
+Input Feature Names
+-------------------
+
+The input feature names are stored in a fitted estimator in a
+``feature_names_in_`` attribute. The feature names are taken from the given
+data during fit. It is taken either from a ``pandas`` data frame or our own
+data structure (something on top of a ``numpy`` array, ``NamedArray`` for
+instance). The attribute will be ``None`` if the given data does not include
+any feature names.
```
NicolasHug: Isn't all this part of the other SLEP? I.e. this is propagation logic which is supposed to be out of scope for this SLEP as indicated above

slep007/proposal.rst
```diff
@@ -0,0 +1,86 @@
...
+
+A fitted estimator exposes the output feature names through the
+``feature_names_out_`` attribute. At the same time, the data structure returned
+by a ``transform`` method also includes the generated feature names.
+
+Same Features Transformers
+**************************
+
+From the perspective of feature names, the simplest transformers are the ones
+which have the same feature names in the output as in the input.
```
NicolasHug: give example of such transformers?

slep007/proposal.rst
```diff
@@ -0,0 +1,86 @@
...
+
+- Input provides feature names: ``feature_names_in_`` and
+  ``feature_names_out_`` are the same.
+- Input provides no feature names: ``feature_names_out_`` will be ``x0`` to
+  ``xn``, where ``n`` is the number of features.
+
+Feature Selector Transformers
+*****************************
+
```
NicolasHug: give example of such transformer?

slep007/proposal.rst
NicolasHug: we can just be conditional here and link to the PR

slep007/proposal.rst
```diff
@@ -0,0 +1,86 @@
...
+***************
+
+Meta estimators can choose to prefix the output feature names or not.
+
+By default, ``Pipeline`` adds no prefix, *i.e* its ``feature_names_out_`` is
+the same as the ``feature_names_out_`` of the last step, and ``None`` if the
+last step is not a transformer.
+
+``ColumnTransformer`` by default adds a prefix to the output feature names,
+indicating the name of the step applied on them.
```
NicolasHug: what about passthrough?

slep007/proposal.rst
adrinjalali: I avoided talking about how this data structure is created in the previous steps in a pipeline, but I thin it's appropriate to talk about how the ``feature_names_in_`` is set in this SLEP.

adrinjalali:
> The SLEP should answer that question and/or address that concern. If we decide to have pca_pca0 and standardscaler_sepallength as feature names that's ok, but it's something that will be produced by the current proposal and that's not mentioned at all.

Now the SLEP says: 

```
This is the default behavior, and it can be tuned by constructor parameters if
the meta estimator allows it. For instance, a ``prefix=False`` may indicate
that a ``ColumnTransformer`` should not prefix the generated feature names with
the name of the step.
```

WDYT?

I also did read through your original PR's description and borrowed quite a bit from there now.

jnothman:
> @thomasjpfan you had a PR where you provided full input-output correspondences, right (or was it a comment somewhere)? 

At one point I proposed a (sparse) matrix provided by each transformer to indicate which inputs corresponded to each output... Another approach would allow us to interrogate precisely how each feature was constructed, but that may be a lot of work for little gain.

slep007/proposal.rst
```diff
@@ -0,0 +1,157 @@
+ .. _slep_007:
+
+=============
+Feature Names
+=============
+
+``scikit-learn`` has been making it easier to build complex workflows with the
+``ColumnTransformer`` and it has been seeing wide adoption. However, using it
```
amueller:
```suggestion
``ColumnTransformer`` and it has been seeing widespread adoption. However, using it
```

slep007/proposal.rst
```diff
@@ -0,0 +1,157 @@
...
+results in very opaque models, even more so than before. There is a `great
```
amueller:
```suggestion
results in very opaque models, even more so than before. There is a `
```

slep007/proposal.rst
```diff
@@ -0,0 +1,157 @@
...
+``female``. However, this is not always possible or desirable especially when a
+generated column is based on many columns, for example in ``PCA``. As a rule of
+thumb, the following types of transformers may generate feature names which
+corresponds to the original features:
+
+- Leave columns unchanged, *e.g.* ``StandardScaler``
+- Select a subset of columns, *e.g.* ``SelectKBest``
+- create new columns where each column depends on at most one input column,
+  *e.g* ``OneHotEncoder``
+- Algorithms that create combinations of ``O(1)`` features, *e.g.*
```
amueller: how is that ``O(1)``? Polynomial is in the name?! it created ``O(n ** degree)`` features ...

slep007/proposal.rst
```diff
@@ -0,0 +1,157 @@
...
+feature names are provided, ``x0`` to ``xn`` are assumed to be their names. For
+example, if a ``SelectKBest`` transformer selects the first and the third
+features, and no names are provided, the ``feature_names_out_`` will be ``[x0,
+x2]``.
+
+Feature Generating Transformers
+*******************************
+
+The simplest category of transformers in this section are the ones which
+generate a column based on a single given column. The generated output column
```
amueller: How is that different from the one-to-one transformers?

slep007/proposal.rst
```diff
@@ -0,0 +1,157 @@
...
+in this case is a sensible transformation of the input feature name. For
+instance, a ``LogTransformer`` can do ``'age' -> 'log(age)'``, and a
+``OneHotEncoder`` could do ``'gender' -> 'gender_female', 'gender_fluid', ...`.
+
+Transformers which generate features based on an ``O(1)`` number of features,
```
amueller: same comment as above. clearly not ``O(1)``

slep007/proposal.rst
```diff
@@ -0,0 +1,157 @@
...
+may generate descriptive names as well. For instance, a
+``PolynomialTransformer`` can generate an output feature name such as ``x[0] *
+x[2] ** 3``.
+
+And finally, the transformers which generate features from a large set of
```
amueller: I don't understand that. I thought the point is more that if an output feature depends on a lot of input features (or all input features) we don't propagate the name.
``PolynomialTransformer`` and ``OneHotEncoder`` typically produce a large number of features, while ``PCA`` produces a small number of features.
The point is that if each output feature depends on all the input features, things will become verbose and/or uninformative.

slep007/proposal.rst
```diff
@@ -0,0 +1,157 @@
...
+
+One-to-one Transformers
+***********************
+
+From the perspective of feature names, the simplest transformers are the ones
+which have the same feature names in the output as in the input, and the
+transformation done on the data is semi-trivial. The ``StandardScaler`` can be
+one example. However, a transformer can choose to be more verbose and generate
+a more informative feature name, ``scaled(income)`` could be an example, and
+the verbosity is controlled by a parameter to the estimator.
```
amueller: later on you suggest ``prefix`` as a parameter name for ``ColumnTransformer``. That's not a very intuitive name. I can't really come up with something better than ``prefix_feature_names``. Would we add this to all transformers?

slep007/proposal.rst
```diff
@@ -0,0 +1,157 @@
...
+last step is not a transformer.
+
+``ColumnTransformer`` by default adds a prefix to the output feature names,
+indicating the name of the step applied on them. If a column is in the output
+as a part of ``passthrough``, it won't be prefixed since no operation has been
+applied on it.
+
+This is the default behavior, and it can be tuned by constructor parameters if
+the meta estimator allows it. For instance, a ``prefix=False`` may indicate
+that a ``ColumnTransformer`` should not prefix the generated feature names with
```
amueller: The example of why always adding this is sometimes ugly is still missing ;)

slep007/proposal.rst
adrinjalali: the one-to-one doesn't change the feature names. I tried to explain that certain transformers may be considered either in the one-to-one or this category, depending on how we decide they should generate names.

slep007/proposal.rst
adrinjalali: tried to clarify

slep007/proposal.rst
adrinjalali: it's about the input features, not the output ones.

slep007/proposal.rst
adrinjalali: Exactly, I don't understand what the question is

slep007/proposal.rst
adrinjalali: sure, it can be `prefix_feature_names`. We don't have to add it to all transformers. Only the ones where it makes sense to tune the generated feature names based on the usecase.

slep007/proposal.rst
amueller: Sorry, I don't see how that clarifies, that's just two conflicting statements. I think the point is that each output feature only depends on a small number of input features. I don't see how polynomial can be O(1).

slep007/proposal.rst
amueller: maybe say "where each output feature only depends on a small number of input features"? Or "where each output feature only depends on a constant number of input features" (which is what O(1)) means.

slep007/proposal.rst
amueller: ok given what you said below it makes a bit more sense. But maybe can you just say that? I.e. each output feature depends on a constant number of input features?

slep007/proposal.rst
amueller: Sorry I don't follow.

slep007/proposal.rst
amueller: Maybe clarify to "transformers where each output feature depends on many or all input features"?

slep007/proposal.rst
amueller: still missing ;)

adrinjalali:
@amueller sorry, hadn't pushed the changes before your latest review.

slep007/proposal.rst
```diff
@@ -49,7 +49,9 @@ corresponds to the original features:
   *e.g* ``OneHotEncoder``
 - Algorithms that create combinations of ``O(1)`` features, *e.g.*
   ``PolynomialFeatures`` on a fixed ``O(1)`` subset of features. The number of
-  output features doesn't have to be ``O(1)``.
+  output features doesn't have to be ``O(1)``. Note that verbosity
```
amueller: I still don't think this is clear. What's wrong with "each output feature depends on a fixed number of input features"?

slep007/proposal.rst
amueller: I think these examples are good, but it's still missing the awkward examples I mentioned.
If you use ``make_column_transformer`` and don't have pipelines in the column tranformer, you'll get
```'onehotencoder_model_100', 'onehotencoder_model_200', ..., 'pca_pca0', 'pca_pca1', ...```
which is probably the much more common usage pattern.
Or you could even get
``standardscaler_num0`` etc

File: slep007/proposal.rst [hunk 21]
```diff
@@ -0,0 +1,223 @@
...
+
+However, the following examples produce a somewhat redundant feature names,
+and hence the relevance of ``prefix_feature_names=False``::
+
+    [model, make, num0, ..., num100] ->
+        make_column_transformer(
+            (OneHotEncoder(), ['model', 'make']),
+            (PCA(n_components=3), ['num0', ..., 'num100'])
+        )
+    feature_names_out_: ['ohe_model_100', 'ohe_model_200', ...,
```
NicolasHug: slightly confused here: where does 'ohe' come from? Shouldn't the step name be `OneHotEncoder`? Same for pca -> PCA

slep007/proposal.rst
```diff
@@ -0,0 +1,223 @@
...
+=============
+
+``scikit-learn`` has been making it easier to build complex workflows with the
+``ColumnTransformer`` and it has been seeing widespread adoption. However,
+using it results in very opaque models, even more so than before. There is a
+`usage example
+<https://scikit-learn.org/dev/auto_examples/compose/plot_column_transformer_mixed_types.html>`_
+in the gallery that applies a classifier to the titanic data set, which is a
+very simple standard usecase.
+
```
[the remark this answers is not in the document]
adrinjalali: does it have to be the case? I think it renders properly, doesn't it?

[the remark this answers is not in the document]
NicolasHug: The issue isn't in the formatting but in the structure of the text which currently makes a new paragraph in the middle of 2 related sentences. 

This makes the structure of the arguments harder to follow (at least for me)

slep007/proposal.rst
```diff
@@ -0,0 +1,223 @@
...
+     'cat__pclass_1',
+     'cat__pclass_2',
+     'cat__pclass_3']
+
+Ideally the generated feature names describe how a feature is generated at each
+stage. For instance, ``cat__sex_female`` shows that the feature has been
+through a categorical preprocessing pipeline, was originally the column
+``sex``, and has been one hot encoded and is one if it was originally
+``female``. However, this is not always possible or desirable especially when a
+generated column is based on many columns, for example in ``PCA``. As a rule of
```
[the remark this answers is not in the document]
adrinjalali: I think adding a justification makes it harder to understand, but i added something.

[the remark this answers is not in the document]
NicolasHug: Yeah I agree, it actually becomes clear later in the ` Feature Generating Transformers ` section.

Feel free to ignore or add a comment saying that

slep007/proposal.rst
```diff
@@ -0,0 +1,223 @@
...
+                         'make_ABC', 'make_XYZ', ...,
+                         'pca0', 'pca1', 'pca2']
+
+Backward Compatibility
+----------------------
+
+All estimators should implement the ``feature_names_in_`` and
+``feature_names_out_`` API. This is checked in ``check_estimator``, and the
+transition is done with a ``FutureWarning`` to give time to third party
+developers to implement the API.
```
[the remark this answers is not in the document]
adrinjalali: The title of the SLEP is feature names, and the API and their generation is discussed here. The other sleps discuss propagation of feature names. Therefore I believe these checks should be mentioned here. The other SLEPs assume the API is already in place.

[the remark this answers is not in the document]
NicolasHug: Maybe then clarify:

- the PR's title which got me confused
- the SLEP title (feature names is too general when other SLEPS also deal with feature names)
- what exactly is the scope of the SLEP in an *abstract* section as done e.g. in SLEPs 9 and 10.

slep007/proposal.rst
```diff
@@ -0,0 +1,223 @@
...
+in this case is a sensible transformation of the input feature name. For
+instance, a ``LogTransformer`` can do ``'age' -> 'log(age)'``, and a
+``OneHotEncoder`` could do ``'gender' -> 'gender_female', 'gender_fluid', ...`.
+
+Transformers where each output feature depends on a fixed number of input
+features may generate descriptive names as well. For instance, a
+``PolynomialTransformer`` on a small subset of features can generate an output
+feature name such as ``x[0] * x[2] ** 3``.
+
+And finally, the transformers where each output feature depends on many or all
```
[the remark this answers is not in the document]
adrinjalali: At the end of the day, it's about how long the generated feature name is, which kinda depends on the number of features which are used to generate each output feature.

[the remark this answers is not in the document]
NicolasHug: Agreed. But then the distinction that is currently made between the 2 above cases is not relevant I think

slep007/proposal.rst
amueller: I still don't follow the distinction between ``LogTransformer`` and ``StandardScaler``. They are exactly the same, right? So either both of them get an "prefix_features_names" parameter or neither would change the names, I'd say. Adding the ``prefix_features_names`` parameter everywhere seems a bit verbose to me.

slep007/proposal.rst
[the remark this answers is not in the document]
amueller: I think the distrinction between O(1) input features per output feature and O(n_features) input features per output feature is ok. I don't see why there's a special case for one-to-one transformers and feature selectors, though. As I said in many of the previous discussions, I see two cases for transformers: combining a fixed number of input features or generating new names.

slep007/proposal.rst
```diff
@@ -0,0 +1,331 @@
...
+input features, generate feature names which has the form of ``name0`` to
+``namen``, where ``name`` represents the transformer. For instance, a ``PCA``
+transformer will output ``[pca0, ..., pcan]``, ``n`` being the number of PCA
+components.
+
+Meta-Estimators
+***************
```
amueller: Shouldn't we list all meta-estimators that are transformers? What about FeatureUnion and RFECV? I guess maybe we're talking about meta-estimators that are not feature selectors, because those are easy.

slep007/proposal.rst
```diff
@@ -0,0 +1,331 @@
...
+
+Meta estimators can choose to prefix the output feature names given by the
+estimators they are wrapping or not.
+
+By default, ``Pipeline`` adds no prefix, *i.e* its ``feature_names_out_`` is
+the same as the ``feature_names_out_`` of the last step, and ``None`` if the
+last step is not a transformer.
+
+``ColumnTransformer`` by default adds a prefix to the output feature names,
+indicating the name of the step applied to them. If a column is in the output
```
amueller:
```suggestion
indicating the name of the transformer applied to them. If a column is in the output
```

slep007/proposal.rst
```diff
@@ -0,0 +1,331 @@
...
+as a part of ``passthrough``, it won't be prefixed since no operation has been
+applied on it.
+
+This is the default behavior, and it can be tuned by constructor parameters if
```
amueller: Why is this restricted to meta-estimators?

slep007/proposal.rst
amueller: ok cool this is added and therefore resolved.

slep007/proposal.rst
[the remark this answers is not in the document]
amueller: I agree, please change the PR title.

amueller:
Actually, I think there is 3 cases for non-meta transformers:
1) Each output feature corresponds to a single input feature. Then just use that input feature.
2) Each output feature is a simple function of a small number of input features. Then the output feature is some representation of the relationship.
3) Each output feature depends on all (or most of) the input features: create new names.

There is some slight tension between 1) and 2) I think. For example if you'd square every entry you'd keep the same feature names, but the entry corresponding to a squared feature in PolynomialFeatures is called x**2. I think 2) and 3) are clear-cut and we are struggling with one a bit. Adding a parameter to all of them would be a possible solution but I'm not sure if it's elegant.

slep007/proposal.rst
adrinjalali: Yeah I know, I was hoping nobody notices that :D I agree they're in principle the same. But to me it makes sense to have different default values for `prefix_feature_names` in different estimators. And I don't think we need it everywhere. So I'm not sure what your concern is.

adrinjalali:
> The main remaining question for me is whether we want to add prefix_feature_names to other transformers as well. It seems verbose but I don't see another way to get log(x0) if that's something that you want.

why does it have to be in all transformers? I thought it makes sense to start with the ones we're not sure what the behavior should be, and whether users want different behaviors or not.

> 1) Each output feature corresponds to a single input feature. Then just use that input feature.
> 2) Each output feature is a simple function of a small number of input features. Then the output feature is some representation of the relationship.
> 3) Each output feature depends on all (or most of) the input features: create new names.

But this misses the `log(x0)` case for a log transformer, for instance; and I think it'd be nice to have the option to have those verbose feature names.

slep007/proposal.rst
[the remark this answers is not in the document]
adrinjalali: Not sure which line this refers to anymore :D setting it to "resolve". Please leave the comment again if it's still an issue :)

slep007/proposal.rst
adrinjalali: in this context it's prefixing it with the name of the underlying transformer.

adrinjalali:
The discussion has become pretty long, and I'm getting lost. Could we please merge this PR, and have the other required changes in different PRs, maybe by you @amueller ? ;)

slep007/proposal.rst
```diff
@@ -0,0 +1,331 @@
...
+=============
+
+``scikit-learn`` has been making it easier to build complex workflows with the
+``ColumnTransformer`` and it has been seeing widespread adoption. However,
+using it results in pipelines where it's not clear what the input features to
+the final predictor are, even more so than before. The following example
+illustrates that very well::
+
+
+    """
```
NicolasHug: Can we trim this down? I know I'm the one who suggested to add the example instead of link to it but I we mostly need the pipeline building, not the full example

adrinjalali: Could we merge it and then you send a PR shortening the example?

NicolasHug: I'd rather not.
Merging PRs with a restricted scope is fine, but merging PRs that are not fully addressed is a slippery slope that I think we should avoid.

NicolasHug:
Please update the title to something less generic than "Feature Names" since other SLEPs are dealing with feature names. The PR title is a great candidate.


Also Please add an abstract section at the top like the previous SLEPs or the template

(Sorry if this is a duplicate, I'm certain to have made this comment yesterday but I don't see it anymore)

slep007/proposal.rst
adrinjalali: we had a chat with @glemaitre and @ogrisel I believe in Paris about this. And to us it seems the SLEP process needs some improvements.

One of the improvements we can think of, is that some of the discussions can happen in separate PRs once the first one is merged. Merging this PR doesn't mean it's accepted or goes for a vote or anything, it just means we can now talk about it in a more structured way. It's not like we have to have only one PR per SLEP.

Also, doing that would make it easier to follow the discussion and the changes. Right now it's really hard to follow the changes.

Another point is that merging them faster would make it easier to distribute the work between us for each SLEP.

adrinjalali:
> Also Please add an abstract section at the top like the previous SLEPs or the template
> 
> (Sorry if this is a duplicate, I'm certain to have made this comment yesterday but I don't see it anymore)

I rather have that in a separate PR by you maybe? ;) I can't think of any good text for that.

slep007/proposal.rst
NicolasHug: I'm open to idea when it's about discussing complex idea (in a sense, that's what you're already doing by creating multiple SLEPs around feature names, and that's fine)

Less-so for simple changes like this one that only take a few minutes to be addressed. Creating a new PR for this will mostly create more work for all of us in general since it will still require a +2.

under_review.rst
```diff
@@ -5,7 +5,7 @@ No SLEP is currently under review.
 
 .. Uncomment below when a SLEP is under review
 
-.. .. toctree::
-..     :maxdepth: 1
+.. toctree::
+    :maxdepth: 1
 
-..     slepXXX/proposal
+    slep007/proposal
```
NicolasHug: please comment the text above that says "No SLEP is currently under review."

slep007/proposal.rst
```diff
@@ -0,0 +1,265 @@
...
+``feature_names_out_`` API. This is checked in ``check_estimator``, and the
+transition is done with a ``FutureWarning`` to give time to third party
+developers to implement the API.
+
+Notes
+-----
+
+This SLEP also applies to `resamplers
+<https://github.com/scikit-learn/enhancement_proposals/pull/15>`_ the same way
+as transformers.
```
NicolasHug: Add this as a `.. note:` in the "scope" section?

slep007/proposal.rst
```diff
@@ -0,0 +1,265 @@
...
+  ``prefix_feature_names`` as explained later can apply here.
+
+This proposal talks about how feature names are generated and not how they are
+propagated.
+
+Scope
+-----
+
+All estimators implement a ``feature_names_in_`` API, and any estimator with a
+``transform`` method implements a ``feature_names_out_`` API, _i.e._ they
```
NicolasHug: Rendering of `_i.e._` is weird I think we need `*i.e.*` here

slep007/proposal.rst
```diff
@@ -0,0 +1,265 @@
...
+  them where there are many. Note that verbosity considerations and
+  ``prefix_feature_names`` as explained later can apply here.
+
+This proposal talks about how feature names are generated and not how they are
+propagated.
+
+Scope
+-----
+
+All estimators implement a ``feature_names_in_`` API, and any estimator with a
```
NicolasHug: I find the term "`feature_names_in_` API" confusing because of the trailing underscore.

I think removing it will disambiguate. Also maybe add e.g.

> By *API*, we mean supporting the attributes, and supporting the features name generation. We discuss these points in the next two sections.

slep007/proposal.rst
```diff
@@ -0,0 +1,270 @@
...
+feature names are generated. Since for most estimators there are multiple ways
+to generate feature names, this SLEP does not intend to define how exactly
+feature names are generated for all of them. It is instead a guideline on how
+they could generally be generated.
+
+Feature Selector Transformers
+*****************************
+
+The output feature names are the ones selected from the input, and if no
+feature names are provided, ``x0`` to ``xn`` are assumed to be their names. For
```
NicolasHug: > if no feature names are provided, ``x0`` to ``xn`` are assumed to be their names.

since this is true for any estimator, we can put this in the intro paragraph of the section above (Output Feature Names). The rest is nice

slep007/proposal.rst
NicolasHug: Marking as resolved since this has been addressed

slep007/proposal.rst
```diff
@@ -0,0 +1,270 @@
...
+Feature Generating Transformers
+*******************************
+
+The simplest category of transformers in this section are the ones which
+generate a column based on a single given column. The generated output column
+in this case is a sensible transformation of the input feature name. For
+instance, a ``LogTransformer`` can do ``'age' -> 'log(age)'``, and a
+``OneHotEncoder`` could do ``'gender' -> 'gender_female', 'gender_fluid',
+...``. An alternative generated feature name for the transformers where each
+output feature corresponds to exactly one input feature is to leave the feature
```
NicolasHug: "An alternative is to leave the feature name unchanged" reads more easily IMHO

slep007/proposal.rst
```diff
@@ -0,0 +1,270 @@
...
+name unchanged. Whether or not to modify the feature name, *e.g.* ``log(x0)``
+vs. ``x0`` is controlled via the ``verbose_feature_names`` to the constructor.
```
NicolasHug: ... may be controlled via a ...

(anticipating the fact that not all estimators need to have the parameter)

slep007/proposal.rst
NicolasHug: Resolving as the parameter is now `verbose_feature_names` and it applies to any estimator.

slep007/proposal.rst
```diff
@@ -0,0 +1,270 @@
...
+
+Output Feature Names
+####################
+
+A fitted estimator exposes the output feature names through the
+``feature_names_out_`` attribute. Here we discuss more in detail how these
+feature names are generated. Since for most estimators there are multiple ways
+to generate feature names, this SLEP does not intend to define how exactly
+feature names are generated for all of them. It is instead a guideline on how
+they could generally be generated.
```
NicolasHug: Maybe add "Furthermore, that specific behaviour of a given estimator may be tuned via the `verbose_feature_names` parameter, as detailed below."

slep007/proposal.rst
```diff
@@ -0,0 +1,270 @@
...
+The API for input and output feature names includes a ``feature_names_in_``
+attribute for all estimators, and a ``feature_names_out_`` attribute for any
+estimator with a ``transform`` method, *i.e.* they expose the generated feature
+names via the ``feature_names_out_`` attribute.
+
+.. note:
+
+    This SLEP also applies to `resamplers
+    <https://github.com/scikit-learn/enhancement_proposals/pull/15>`_ the same
+    way as transformers.
```
NicolasHug: Somehow this isn't rendered. Maybe just remove the `.. note:`, it's fine as regular text.

slep007/proposal.rst
```diff
@@ -0,0 +1,270 @@
...
+
+
+    [f1, ..., f100] -> Pipeline(
+                           [SelectKBest(k=30),
+                            PCA(n_components=3)]
+                       )
+    feature_names_out_: [pca0, pca1, pca2]
+
+
+    [model, make, num0, ..., num100] ->
```
NicolasHug: Id suggest numeric0 ... numeric100
instead of
num0 ... num100

to clear our any potential confusion with the name of the CT step which is also 'num'

slep007/proposal.rst
```diff
@@ -0,0 +1,270 @@
...
+    feature_names_out_: ['model_100', 'model_200', ...,
+                         'make_ABC', 'make_XYZ', ...,
+                         'pca0', 'pca1', 'pca2']
+
+Backward Compatibility
+######################
+
+All estimators should implement the ``feature_names_in_`` and
+``feature_names_out_`` API. This is checked in ``check_estimator``, and the
+transition is done with a ``FutureWarning`` to give time to third party
```
NicolasHug: for at least 2 versions

glemaitre:
OK, I will catch up with all the discussion and then merge such that we can open subsequent PR on some specific points of the SLEP.

NicolasHug:
Why? We can address the remaining points here, if any. The SLEP is reasonably small (and in good shape now) and I'd rather be reviewing one PR than many

glemaitre:
> The SLEP is reasonably small (and in good shape now) and I'd rather be reviewing one PR than many.

I have to go across 133 messages and unstack. For me, this is exactly the point raised by @adrinjalali there https://github.com/scikit-learn/enhancement_proposals/pull/17#discussion_r377676740 which make it difficult to catch up with SLEP that I did not look at since a while.

NicolasHug:
In the case of this PR there's only one discussion that hasn't been resolved in the diff.

> I have to go across 133 messages and unstack

I agree this is a problem. Personally, I try to provide timely feedback when I'm reviewing, and also try to address comments fast when I'm the author. It doesn't always work out though. Ideally, we would prioritize PRs better and dedicate time, e.g. during the monthly meetings.

I'm being grumpy but I think opening new PRs will mostly scatter discussions, making it even more difficult to catch up. And it may add more work for everyone. But OK, let's try for the SLEPs. I don't think we should be doing that for main repo though.

slep007/proposal.rst
```diff
@@ -0,0 +1,272 @@
...
+:Type: Standards Track
+:Created: 2019-04
+
+Abstract
+########
+
+This SLEP proposes the introduction of the ``feature_names_in_`` attribute for
+all estimators, and the ``feature_names_out_`` attribute for all transformers.
+We here discuss the generation of such attributes and their propagation through
+pipelines.
```
glemaitre: I would make it obvious from this point something that you later introduce: "Since for most estimators there are multiple ways to generate feature names, this SLEP does not intend to define how exactly feature names are generated for all of them". Basically, I only think it would be great to stipulate in the abstract that we don't give all implementation details but rather a guideline.

slep007/proposal.rst
```diff
@@ -0,0 +1,272 @@
...
+A fitted estimator exposes the output feature names through the
+``feature_names_out_`` attribute. Here we discuss more in detail how these
+feature names are generated. Since for most estimators there are multiple ways
+to generate feature names, this SLEP does not intend to define how exactly
+feature names are generated for all of them. It is instead a guideline on how
+they could generally be generated. Furthermore, that specific behavior of a
+given estimator may be tuned via the ``verbose_feature_names`` parameter, as
+detailed below.
+
+When generating the output feature names, if no input feature names are
```
glemaitre: Just a thought but I don't have a better proposal (so feel free to skip it). I find this sentence a bit lonely in this section. However, it does not fit in other section so I don't know how to make it better.

slep007/proposal.rst
```diff
@@ -0,0 +1,272 @@
...
+the same as the ``feature_names_out_`` of the last step, and ``None`` if the
+last step is not a transformer.
+
+``ColumnTransformer`` by default adds a prefix to the output feature names,
+indicating the name of the transformer applied to them. If a column is in the output
+as a part of ``passthrough``, it won't be prefixed since no operation has been
+applied on it.
+
+This is the default behavior, and it can be tuned by constructor parameters if
+the meta estimator allows it. For instance, a ``verbose_feature_names=False``
```
glemaitre: After reading all the SLEP, one thing which is not clear is what `verbose_feature_names` can and cannot be. After the first occurrence, I thought that it could a sort of callable to fine-tune the naming. In the remaining of the document, it look like this is only a boolean.

I don't know if we could be a bit more precise on what `verbose_feature_names` could be or at least mention which extension we can expect.

slep007/proposal.rst
```diff
@@ -0,0 +1,288 @@
...
+
+verbose_feature_names
+*********************
+
+``verbose_feature_names`` controls the verbosity of the generated feature names
+and it can be ``True`` or ``False``. Alternative solutions could imagine:
+
+- an integer: fine tuning the verbosity of the generated feature names.
+- a ``callable`` which would give further flexibility to the user to generate
+  user defined feature names.
```
NicolasHug: simpler than a callable would be to accept a format string with a specific language, i.e.

users may pass `"wrapping_with_parenthesis({input_name})"`

and one our side we would do `"wrapping({input_name})".format(input_name=whatever_the_column_name_is)`

so that the final name would  be `wrapping_with_parenthesis(f0)`

A given string could actually be the default, to examplify the use

NicolasHug: Which would change the parameter name to `names_format` whose default could be e.g. `"log({input_name})"` for the LogTransformer or `"pca_{input_name}"` for PCA

adrinjalali: yes, we can talk about alternatives, but this SLEP says we're not implementing any of them for now anyway.

slep007/proposal.rst
```diff
@@ -0,0 +1,288 @@
...
+  ``verbose_feature_names`` as explained later can apply here.
+
+This proposal talks about how feature names are generated and not how they are
+propagated.
+
+verbose_feature_names
+*********************
+
+``verbose_feature_names`` controls the verbosity of the generated feature names
+and it can be ``True`` or ``False``. Alternative solutions could imagine:
```
NicolasHug:
```suggestion
and it can be ``True`` or ``False``. Alternative solutions could include:
```

adrinjalali: I changed it, but this is the language I liked and enjoyed, and I don't understand why you'd change it you understand what the text means.

NicolasHug: I'm not trying to be an ass @adrinjalali .

I'm not an English native speaker and my understanding is influenced by my mother tongue: French. In French, only a living being can "imagine" something. The current sentence says that a "solution" may "imagine" something.

I believe it's incorrect in English, so I'm suggesting an alternative. I might be wrong.

adrinjalali:
Seems we're mostly happy with the current state of this SLEP. I'll merge to have it on the website, then wait a bit for more feedback, and hopefully soon call a vote on it :)

NicolasHug:
This PR was merged with only one approval (mine), and with an unaddressed comment from @amueller https://github.com/scikit-learn/enhancement_proposals/pull/17/files#r373808675. How and when will this comment be addressed? I don't believe it's less worthy than other comments.

adrinjalali:
1. Forgetting about that is one of the reasons I'd like to merge SLEPs much faster, to keep those discussions contained in their own issue/PR.
2. This is now merged, not accepted, we can continue the discussion on separate issues. I'll open an issue with that question.

NicolasHug:
Thanks.

Just a tip, it might seem obvious to some but I know that it's not always the case (for having discussed with other devs): looking at the diff (https://github.com/scikit-learn/enhancement_proposals/pull/17/files) is a great way to see which comments are still unaddressed, compared to the main conversation page. There might still be comments in the conversation page but at least those in the diff are hard to miss.

slep007/proposal.rst
amueller: I'm pretty sure the previous version was not grammatical, I thought it was a typo.

PR #27, opened 2020-01-12
[MRG]  SLEP010 from "under review" to "accepted"

NicolasHug:


lorentzenchr:
Vote in https://mail.python.org/pipermail/scikit-learn/2019-December/003553.html.
Do you think, one should link the vote on the website as is done for PEPs?

PR #28, opened 2020-01-30
SLEP011 Do not make mandatory TC vote during SLEP vote

glemaitre:


slep011/proposal.rst
```diff
@@ -0,0 +1,57 @@
...
+:Status: Draft
+:Type: Standards Track
+:Created: 2020-01-30
+
+Abstract
+########
+
+This SLEP proposes a change in the decision making process. It proposes to not
+always call for a TC vote in case that no consensus is found during the vote of
+a SLEP. Instead, the TC should be appeal by the SLEP proposer.
```
adrinjalali:
```suggestion
a SLEP. Instead, the matter goes to the TC if the proposer appeals to the TC.
```

Maybe clearer?

slep011/proposal.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Motivation
+##########
+
+Currently, the decision making process included in the governance document
+include the following statement:
```
adrinjalali:
```suggestion
includes the following statement:
```

adrinjalali:
For instance, with the resampler SLEP, there needs to be two SLEPs proposing the two solutions, and if one of the is rejected, we don't necessarily want the TC to have to be involved, the slep author can go and modify or have a new one or something. Discussing what will happen in the case of the sleps we already have yesterday, we realized this is an issue which needs to be fixed.

slep011/proposal.rst
```diff
@@ -0,0 +1,57 @@
...
+[...]"
+
+It might not be necessary to call for a TC vote if the proposer want to review
+the SLEP and make a new proposal.
+
+Solution
+########
+
+"[...] If no option can gather two thirds of the votes cast, the decision is
+escalated to the TC **if requested by the SLEP proposer**, which in turn will
```
GaelVaroquaux:
```suggestion
escalated to the TC **if the SLEP proposer would like the SLEP to be accepted**, which in turn will
```

GaelVaroquaux:
I'm overall fine with the principle of this suggestion. It is a good one!

amueller:
Not sure if this is our most urgent issue but I'm also ok with it

adrinjalali:
@glemaitre could you also add this to the under review maybe? we can then merge and call the vote.

slep011/proposal.rst
```diff
@@ -0,0 +1,57 @@
...
+[...]"
+
+It might not be necessary to call for a TC vote if the proposer want to review
+the SLEP and make a new proposal.
+
+Solution
+########
+
+"[...] If no option can gather two thirds of the votes cast, the decision is
+escalated to the TC **if the SLEP proposer would like the SLEP to be accepted**, which in turn will
```
NicolasHug: Line too long

amueller:
The current governance document doesn't require a vote if there is consensus. We have voted on previous SLEPs but there's not really any need to vote for this one, I feel.

adrinjalali:
> The current governance document doesn't require a vote if there is consensus. We have voted on previous SLEPs but there's not really any need to vote for this one, I feel.

That's an interesting point, which makes it possible to make important decisions w/o announcing on the mailing list beforehand. Might be a bug, not a feature :P

amueller:
You think it is? Well then we should... probably fix that first lol?
Add a sentence "before any SLEP is accepted it should be announced on the mailing list"?

amueller:
Actually, rereading, I think the governance doc is not clear on whether a vote is required :(
I thought the intent was to require a vote only if there is disagreement.

GaelVaroquaux:
> That's an interesting point, which makes it possible to make important decisions w/o announcing on the mailing list beforehand. Might be a bug, not a feature :P

Hum, yes.

amueller:
@GaelVaroquaux did you understand the governance document as always requiring a vote for a SLEP?

rth:
As a side comment, making votes for SLEPs on the mailing list, especially when there is unanimous agreement and all discussion happened in the PR is probably not very relevant for most subscribed users. For most I imagine it's just noise, which will not help in keeping users subscribed/interested. 

I get that we need some public and temper proof way to record votes, but I wonder if the user mailing list is the best place for it.

GaelVaroquaux:
> @GaelVaroquaux did you understand the governance document as always requiring a vote for a SLEP?

No I did not, but the point that you raise about SLEPs slipping below the
radar for lack of communication is a good one.

adrinjalali:
This is an alternative path to reduce noise on the mailing list:

- merge a SLEP as under review
- create a PR to move the SLEP as accepted
- send an email, and ask core-devs to vote on the PR

slep011/proposal.rst
```diff
@@ -0,0 +1,57 @@
...
+:Author: Guillaume Lemaitre
+:Status: Draft
+:Type: Standards Track
+:Created: 2020-01-30
+
+Abstract
+########
+
+This SLEP proposes a change in the decision making process. It proposes to not
+always call for a TC vote in case that no consensus is found during the vote of
```
jorisvandenbossche:
```suggestion
always call for a Technical Committee (TC) vote in case that no consensus is found during the vote of
```

glemaitre:
I made the changes. I think the proposal of @adrinjalali is what I would expect.
We should probably fix these issues as well. What is the path then?

adrinjalali:
I don't think we need to change anything since the governance model doesn't really specify how it should be done. I'd say we merge, create the next PR to move it as accepted, and call the vote on the mailing list.

amueller:
So should we merge this? Or 000?

GaelVaroquaux:
With regards to merging this SLEP, I fear that, in it's current state it would add an extra layer of stuff to read. In other terms, I think that we need to consolidate the information somewhere, elsewhere people will have to dig for this information through many layers. Where is that somewhere, I am not certain. Maybe SLEP 000 is a good way to doing that.

glemaitre:
So let's focus on SLEP000 and come back once we have something solid.

GaelVaroquaux:
I think that this is a pragmatic view! :)

Issue #35, opened 2020-02-18
SLEP012 input_feature_names_


thomasjpfan:
In SLEP012, it suggests a new attribute, `input_feature_names_`, onto the estimators:

https://github.com/scikit-learn/enhancement_proposals/blob/master/slep012/proposal.rst#L25

Is this in scope of the SLEP?

adrinjalali:
I kinda wanted to have it about the `DataArray` only at first, but through discussions we decided to include that in the SLEP.

thomasjpfan:
It looks like the two motivations for the `InputArray` is for:

1. Input feature names for model inspection, which pushes for defining `input_feature_names_`.
2. Allow estimators to treat columns differently. (For fairness or categories)

Would voting for SLEP 012 indirectly vote for `input_feature_names_`? If not, the SLEP vote is kind of for "infrastructure". In other words, it will enable us to work on the two items above.

(BTW I trying to see what we need to do to get SLEP 012 to be ready to vote on)

Is there any other use cases you see for `InputArray`?

adrinjalali:
I guess I need to give a bit of a context.

The history of feature names "feature" (:D) seems something like the following to me (please correct me if I'm missing something @amueller and @jorisvandenbossche ):

1. a `get_feature_names` is introduced in some of the estimators. In most estimators it's the _output_ feature names, but if implemented in a predictor, or a pipeline which has a predictor as the last step, it's the input features to the predictor (hence kind of inconsistent).
1. Propagating the feature names through a pipeline in order to be able to inspect the feature names going into each step, can be done if we allow `get_feature_names` to accept an existing set of feature names (from the previous step, in the pipeline), and having the `get_feature_names` of the pipeline doing that for us, would allow the user to inspect the features going to the last step. This would be done once the `fit` is called. Some of the drawbacks: `get_feature_names` not clear if it's the input or output, only works in a pipeline, estimators don't have access to the feature names at `fit` time (@amueller has an implementation of this).
1. Pipeline propagates the feature names at fit time, getting the output feature names of each step, passing it to the next step's `fit`. This could either be done with `get_feature_names` or the `feature_names_in_` and `feature_names_out_`. The latter makes it clear which one's which. It also means estimators have access to the feature names during `fit`. However, it's quite a bit of change in the pipeline, and also requires us to pass an especial `feature_names` arg to `fit`, which doesn't seem the nicest design (@amueller has an implementation of this).
[...]

PR #37, opened 2020-02-19
SLEP 014 Pandas in Pandas out

thomasjpfan:
This SLEP proposes pandas in pandas out as an alternative to [SLEP 012 InputArray](https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep012/proposal.html).


NicolasHug:
I think this should be slep 14

slep014/proposal.rst
```diff
@@ -0,0 +1,173 @@
...
+wrap the ndarray into a ``DataFrame`` as a return value. This is will be
+piped into ``LogisticRegression.fit`` which calls ``check_array`` on the
+``DataFrame``, which may lead to a memory copy in a future version of
+``pandas``. This leads to unnecessary overhead from piping the data from one
+estimator to another.
+
+Backward compatibility
+######################
+
+The ``set_config(pandas_inout=True)`` global configuration flag will be set to
```
adrinjalali: maybe more like `pandas_output`?

adrinjalali:
Another point is to talk abut `pandas` being a soft dependency in the SLEP, I guess; I'm not sure if we need to talk about how.

TomAugspurger:
> One of the reasons I think we should consider xarray, is that we can attach arbitrary feature/sample/data props to the data

pandas 1.0 added a `DataFrame.attrs` dictionary that behaves the same as DataArray.attrs, so xarray and pandas should be the same in this regard.

> and we may be able to convince them to remove a hard pandas dependency (correct me if I'm wrong @TomAugspurger)

In theory, I think it's possible to have a `Variable` (which backs a DataArray) without any coordinates (and so no need for a pandas Index to store the labels). I don't have a good idea of how open the xarray devs are to extracting that / making it possible to use xarray without pandas.

slep014/proposal.rst
thomasjpfan: I had something like that initially, but it feels like "We will always output pandas, even if the input is numpy arrays". (Naming is hard)

slep014/proposal.rst
```diff
@@ -0,0 +1,173 @@
...
+    set_config(store_feature_names_in=True)
+
+    pipe.fit(X, y)
+
+    pipe['logisticregression'].feature_names_in_
+
+3. Allow for extracting the feature names of estimators in meta-estimators::
+
+    from sklearn import set_config
+    set_config(store_feature_names_in=True)
```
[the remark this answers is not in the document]
thomasjpfan: The default of `pandas_inout` is now stated twice in the Solution and Backward compatibility sections. The default value of `store_feature_names_in` is started in its section. I purposefully did not go into too many details in the **Enabling Functionality** section, since it serves as "what are the possibilities if this SLEP gets accepted".

thomasjpfan:
> Should we consider the new pandas nan in the consideration section? I think if transformers return pandas dataframes, it becomes relevant.

The `pd.Na` feature is consider experimental. As [stated in their docs](https://pandas.pydata.org/docs/whatsnew/v1.0.0.html#experimental-na-scalar-to-denote-missing-values):

> Experimental: the behaviour of pd.NA can still change without warning.

It also seems like [DataFrame.attrs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.attrs.html?highlight=attrs) is experimental as well.

In both cases, I think we should wait for the features to stabilize before designing with them in mind.

slep014/proposal.rst
```diff
@@ -0,0 +1,197 @@
...
+For options 2 and 3 the default value of configuration flag:
+`store_feature_names_in` is False.
+
+Considerations
+##############
+
+Index alignment
+---------------
+
+Operations are index aligned when working with DataFrames. Interally,
```
jorisvandenbossche:
```suggestion
Operations are index aligned when working with DataFrames. Internally,
```

slep014/proposal.rst
```diff
@@ -0,0 +1,197 @@
...
+    set_config(pandas_inout=True)
+    pipe = make_pipeline(StandardScaler(), LogisticRegression())
+    pipe.fit(X, y)
+
+Interally, ``StandardScaler.fit_transform`` will operate on a ndarray and
+wrap the ndarray into a DataFrame as a return value. This is will be
+piped into ``LogisticRegression.fit`` which calls ``check_array`` on the
+DataFrame, which may lead to a memory copy in a future version of
+``pandas``. This leads to unnecessary overhead from piping the data from one
+estimator to another.
```
jorisvandenbossche: Personally, I think that for some transformers (like StandardScaler) could rather easily work column-wise to avoid such copying overhead. 
(of course, that will give a stronger pandas dependence also for the implementation of such transformers)

jorisvandenbossche:
> Should we consider the new pandas nan in the consideration section? I think if transformers return pandas dataframes, it becomes relevant.

I personally think it's interesting that scikit-learn would consider supporting it, but IMO it is somewhat orthogonal to this SLEP. Also in the current situation of converting DataFrames to arrays on input to estimators, the question of supporting pd.NA already comes up.

amueller:
I think the scope is only ``transform``. If that's not clear from the SLEP it needs to be clarified. Or are you saying it should also be ``predict`` etc? I'm not sure how I feel about that ;)

jorisvandenbossche:
I think transforms is the most important to start with, so that seems the best scope. But yes, so I mainly wanted to indicate that this should be mentioned more clearly in the SLEP then ;)

slep014/proposal.rst
```diff
@@ -0,0 +1,197 @@
...
+    from sklearn import set_config
+    set_config(pandas_inout=True)
+
+    pipe.fit(X, y)
+    X_trans = pipe[:-1].transform(X)
+
+    X_trans.columns.tolist()
+    ['letter_a', 'letter_b', 'letter_c', 'pet_dog', 'pet_snake', 'num']
+
+This introduces a soft dependency on pandas, which is opt-in with the
```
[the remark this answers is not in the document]
adrinjalali: So maybe word it more like

> This feature is only available through a soft dependency on `pandas`

or something.

slep014/proposal.rst
adrinjalali: and it could even have the option of being "in-place" :D

thomasjpfan: We can try to support this. `check_array` would need to not run `asarray` on the dataframe, and the transformer would need to operate on the dataframe itself.

amueller: yes ``check_array`` could get another option! exciting! ;)

slep014/proposal.rst
```diff
@@ -0,0 +1,202 @@
...
+############
+
+- :ref:`SLEP012 Custom InputArray Data Structure <slep_012>`: This approach
+  adds another data structure in the Python Data ecosystem. This increases
+  the maintenance responsibilities of the ``scikit-learn`` library.
+
+- Use xarray's Dataset, ``xr.Dataset``: The pandas DataFrame is more widely used
+  in Python's Data ecosystem, which means more libraries are built with pandas
+  in mind. With xarray support, users will need to convert their DataFrame into
+  a ``xr.Dataset``. This converstion process will be lossy when working with
```
glemaitre:
```suggestion
  a ``xr.Dataset``. This conversion process will be lossy when working with
```

amueller:
Do we want to have an in-depth discussion of xarray vs pandas in this slep? In this issue? In #35?

My brief opinion:
### pandas pros
- allows heterogeneous data types
- is the natural format for tabular data (categorical dtypes, lots of io stuff, widely used)
### pandas cons
- potential memory copy for round-trip in the future

### xarray pros
- guaranteed no memory copy for round-trip
### xarray cons
- homogeneous data only
- we need to come up with names for the axes and users need to adhere to them
- adds one additional dependency

I assume @adrinjalali's pro and con list is different from mine since his preference is xarray ;)

glemaitre:
> xarray cons
> * homogeneous data only

Isn't it possible to have it with an xarray `Dataset`?

amueller:
That's true. I thought we were considering ``DataArray`` not ``DataSet``. If we're using ``DataSet`` we have to basically reimplement the pandas block manager, right?

glemaitre:
> If we're using DataSet we have to basically reimplement the pandas block manager, right?

I don't know enough about it but it seems that you are right.

amueller:
I also don't know much about it, that was an educated guess based on the docstring ;)

glemaitre:
Same educated guest then :)
 

                                                                                                                                      
 

 
     

Sent from my phone - sorry to be brief and potential misspell.

jnothman:
Is the copying issue in pandas a real one? I would expect that we could petition pandas to make the data F-contiguous and export a non-copied values array, in the case that the frame is being constructed anew with a homogeneous dtype. This would be the case for almost all of our transformer outputs. 

Do we need to choose One format? It seems that a big challenge in this feature altogether is trying to support different formats (non ndarray) and their proliferation. Should we be providing for adapters to support different types?

PS: I've not read much of the above

amueller:
> Is the copying issue in pandas a real one? 

It's unclear to me at least. I talked to @WillAyd a couple of days ago and he seem to think a zero-copy round-trip is feasible even when moving to a column store.

> Do we need to choose One format?
We don't need to choose "One Format" but for each input type, we need to choose one output format, also see https://github.com/scikit-learn/enhancement_proposals/pull/25#issuecomment-571640817.

And I guess then the question is which ones should we implement.

> Should we be providing for adapters to support different types?
Are adapter meta-estimators? I think that would introduce too much complexity. They could be a stop-gap but I'd rather have a better solution.

So far I don't think any of the proposed duck-array protocols would help us here. If we want to support arbitrary types, we'd need a protocol that basically allows us to separate the numpy array from the meta-data and then recreate a new duck array with a new numpy array and the old meta-data. I'm not entirely sure in how far that's possible, and it seems a little bit overkill to me for now.

jorisvandenbossche:
> a zero-copy round-trip is feasible even when moving to a column store.

If you want a 2D numpy array, then a zero-copy roundtrip is not possible with a column store AFAIK.

amueller:
I'm not saying that it's a good idea, but in theory one could find out that the columns are sequential in memory and make a view of them together, right?
(you'd probably also keep around the separate views so no-one deallocates it)

```
import numpy as np
X = np.random.normal(size=(100, 3)).copy("F")

asdf = {'a': X[:, 0], 'b': X[:, 1], 'c': X[:, 2]}

asdf['b'].__array_interface__['data'][0] == asdf['a'].__array_interface__['data'][0] + asdf['a'].itemsize * asdf['a'].size
```
> True

lorentzenchr:
Another 5 cents to this discussion.

In my view, there are 2 important design choices for a ML library concerning data structures:
1. Internal data structure: What kind of object is passed inside a pipeline, from step to step? So far, this is a numpy array, a homogeneous data structure.
2. Connectivity to other data structures at the very beginning of a pipeline. So far, you can pass a pandas dataframe (heterogeneous), e.g. to a column transformer. In the first step, it is converted to the internal data structure.

This SLEP is about the first point, the internal data structure, and therefore touches the very foundation and goes far beyond feature names (at least in this pandas in pandas out version). This brings me to the following (provocative<sup>1</sup>) questions:
- If all estimators swallow (only) numpy arrays, why aren't they in scipy? Example: linear models.
- There are estimators like tree-based ones that could easily deal directly with heterogeneous data structures (as long as each column can be sorted). Therefore, would it make sense to use a heterogeneous internal data structure?
- Which internal data structure is the best one for
   - ease of use (I want to use column names, sample props, ...)
   - computational efficiency
   - memory efficiency (does the input data stay untouched?)
   - compatibility to and dependency on other libraries and frameworks
- If you were to build scikit-learn again next year, which solution would you chose?

<sup>1</sup> So some questions better stay unanswered:smirk:

thomasjpfan:
From my point of view, we can not really separate the data structure into internal and external (in terms of being in a pipeline). When a third party constructs a custom transformer, `transform` would need to output a data structure based on a configuration flag.

The most memory efficient thing to do is to have our own data structure, where we have complete control of it (`InputArray` SLEP)

thomasjpfan:
> This SLEP is about the first point, the internal data structure, and therefore touches the very foundation and goes far beyond feature names

One of the motivation goals is to get feature names. This SLEP describes an approach to get there and what the ramifications are for this approach.

From my point of view, we can not really separate the data structure into internal and external (in terms of being in a pipeline). When a third party constructs a custom transformer, `transform` would need to output a data structure based on a configuration flag. In all cases, we will be putting more requirements on third party estimators.

The most efficient thing to do is to have our own data structure, where we have complete control of it (i.e. `InputArray` SLEP). Users would need to convert this data structure into a pandas dataframe to be able to use it with other libraries. Since this is a new data structure, third parties will need to learn about how to construct and use it. 

This SLEP focuses on using pandas, which most developers in the PyData ecosytem knows how to use. The biggest downside is that we do not have control over it and it will most likely not be as efficient.

A mixture between the two approaches would be great:

1. When a transformer is in a pipeline, output the internal data structure so we can optimize it.
2. When the transformer is by itself, output a pandas dataframe so end users can use it with their favorite framework.

We would create helper functions for developers so they can more easily support this.

amueller:
We could distinguish between optimized internal and external and it's something I have thought about. I don't think we really gain a lot by making this distinction as it would complicate the interface with a negligible impact on performance. We would still have to implement the "external" interface within the pipeline to support third-party estimators, as @thomasjpfan says.
 
I don't really think this is about efficiency, it's about interfaces. As I said, I'm not really concerned with a potential copy right now. But even if we consider that, I don't think there's a clear answer to what the "most efficient" approach would be. 
The "most efficient" data structure for OneHotEncoding, gradient boosting and logistic regression might also not be the same.

I'm answering some of @lorentzenchr's questions below.

<details>
- If all estimators swallow (only) numpy arrays, why aren't they in scipy? Example: linear models.

Why should they be? Ideally scipy should be split up. The only reason it exists the way it does is because packaging code with fortran dependencies used to be hard before we had conda and wheels. Today, I hope no-one would create a monster like that.

- There are estimators like tree-based ones that could easily deal directly with heterogeneous data structures (as long as each column can be sorted). Therefore, would it make sense to use a heterogeneous internal data structure?

That is likely going to be quite inefficient. What is best in the trees is to use binning and then everything will be an integer which allows using homogeneous arrays and everything will be very efficient. Right now pandas doesn't have a C interface, so it's basically a non-starter (though it might get one at some point). Even if there was a C interface, dealing with variable column types in a strongly typed language is likely to be an inefficient nightmare.

- If you were to build scikit-learn again next year, which solution would you chose?

With a global config flag we can make relatively arbitrary backward-incompatible changes. So we're not really bound to anything and that is the main question I think we're trying to answer.

</details>

adrinjalali:
I quite agree with @amueller 's assessment on internal vs external data structures. I really don't think we should go down that road.

I had another talk with @TomAugspurger , and I think:

Whatever we choose, should stay there for a while. This is kind of a big change and I don't think it'd be nice to introduce the change, and then realize we should change again in two years, and this is why I would consider the upcoming changes in pandas into account when considering the options.

That said, there are two aspects which make me be more in favour of xarray vs pandas. One is the changes which [most probably] coming to pandas, the other one is that xarray is much more lightweight and much closer to numpy arrays, hence less change would be required from our side.

On the change side, we have the columnar representation which would mean two guaranteed copies on the `df.DataFrame(np.asarray(dataframe))` operation, even if they're all `float`s. I do believe this is not a desired behavior if the user wants to have feature names.

Another change is the `NA` vs `nan` distinction coming to pandas. I do think it'd be nice to distinguish between the two, i.e. one missing and one the output of `0/0` for instance, but the fact that numpy doesn't have support for missing values, would mean further distance between what operations should do on those values. For instance, would the imputer then fill the `NA` values but not `nan`s since `nan`s are not missing values anymore? It may be a good idea, but I'm not sure if we want to go down that road now. Also, we could have the support in imputers for them to understand the distinction and accept a pandas DataFrame, but I'm not sure if we'd want to expand that to the rest of the library.

On the other hand, xarray data structures are pretty much numpy arrays with some metadata attached to them. It'd be much easier for us to work with them since we've been doing it all along, and not much changes. It's also pretty lightweight, and my gut feeling is that they'd be open to some changes if needed be.

Another aspect is that xarray containers have been supporting other metadata than just the feature/row names, and it may be useful for us to attach them to the data (another way to do sample/feature/data props for instance).

When I think of these two data structures, I really feel the xarray ones are closer to what we'd like to have and work with. And to be clear, by no means I'm saying users should give us xarray containers. They can, but the way I imagine the workflow is:

``` python
df = fetch_openml(..., as_frame=True)
# do whatever needed with the dataframe

pipe = Pipeline(...).fit(df) # predictor pipeline
pipe.predict(df) # works and feature names are all set and passed through

Xt = Transformer().fit(df).transform(df) # returns an xarray
df_t = sklearn.utils.dataframe(Xt) # returns a pandas DataFrame
```

Also, I think there's value in supporting the `xarray.Dataset` for heterogeneous data. But need to experiment with it a bit further.

thomasjpfan:
@adrinjalali In the imagined workflow, it seems like the transformer needs to understand pandas and xarray as input. Specifically, the output of the transformer is an xarray and if it is placed in a pipeline, the next transformer would need to take it as input.

This means third party estimators would need to do, xarray or pandas in -> xarray out?

TomAugspurger:
Just my bias, but I'd like to see `type(X) in -> type(X) out`, at least for ndarray, pandas.DataFrame, and xarray.DataArray (possibly xarray.Dataset, not sure). I'm not sure if that makes sense for every transformer, but if it worked I think people would be happy. I think it also satisfies some of Adrin's concerns.

> On the change side, we have the columnar representation which would mean two guaranteed copies on the df.DataFrame(np.asarray(dataframe)) operation, even if they're all floats. I do believe this is not a desired behavior if the user wants to have feature names.

If / when pandas switches to a column store (which isn't guaranteed at this point), users will face that double-copy when providing a DataFrame at the start of the pipeline. But if both DataFrame's and xarray objects are accepted, users can reduce that to a single copy up front with a `DataFrame.to_xarray()` at the start of the pipeline.

> Another change is the NA vs nan distinction coming to pandas.

I think this is mostly orthogonal to the container discussion. This can happen today when a user provides a DataFrame with NAs, and scikit-learn converts it to an object-dtype ndarray (which will presumably cause errors).

> Another aspect is that xarray containers have been supporting other metadata than just the feature/row names, and it may be useful for us to attach them to the data (another way to do sample/feature/data props for instance).

Pandas recently added `.attrs`, which mirrors xarray's and H5py's. It's currently experimental but hasn't had any bug reports, so we'll move it to stable soon. I think transformers can reasonably rely on these behaving the same between pandas and xarray.

amueller:
First, I think it's important to capture all these arguments in a SLEP, so they are nicely summarized.

Second, I think @thomasjpfan makes an important point, which also was my first thought reading your message: we need to distinguish input and output. I agree that handling pandas input does make things tricky (I think more so than pandas output). However, I think we do want to handle pandas input. So using xarray as output will not really save us from that.
@TomAugspurger made this great table in https://github.com/scikit-learn/enhancement_proposals/pull/25#issuecomment-571640817
( I think the table has a typo though, the last row appears twice and ndarray input doesn't appear for dataframe fit)

> On the change side, we have the columnar representation which would mean two guaranteed copies on the df.DataFrame(np.asarray(dataframe)) operation, even if they're all floats. I do believe this is not a desired behavior if the user wants to have feature names.

Shouldn't that be only a single copy?


To a first approximation, I think we can ignore the input to fit to make the table a bit easier, and look at transform input vs transform output.
So @adrinjalali's proposal is

Transform Input | Transform Output
--------------- | ----------------
ndarray         | ndarray
DataArray       | DataArray
DataFrame       | DataArray
AnythingElse    | ndarray

@TomAugspurger's proposal is

Transform Input | Transform Output
--------------- | ----------------
ndarray         | ndarray
DataArray       | DataArray
DataFrame       | DataFrame
AnythingElse    | ndarray


and I my proposal was

Transform Input | Transform Output
--------------- | ----------------
ndarray         | ndarray
DataFrame       | DataFrame
AnythingElse    | ndarray

amueller:
@adrinjalali can you maybe rephrase your arguments in terms of how xarray is better for outputs? And/or argue why pandas special behaviors are not an issue for inputs?

amueller:
> This means third party estimators would need to do, xarray or pandas in -> xarray out?

They could always fallback to ndarray for either of them and things would "work" but it would swallow feature names and other meta-data.

TomAugspurger:
> Shouldn't that be only a single copy?

Too in the weeds for this discussion, but it's at least 1, possibly two (depending on if we automatically go to Arrow memory 😄). Safe to say it's one or more copies.

> They could always fallback to ndarray for either of them and things would "work" but it would swallow feature names and other meta-data.

That's the hardest part for me... It essentially tying feature names to the ability of a transformer to return a DataFrame (or DataArray). Is that an acceptable limitation for scikit-learn?

amueller:
> That's the hardest part for me... It essentially tying feature names to the ability of a transformer to return a DataoFrame (or DataArray). Is that an acceptable limitation for scikit-learn?

I think it is acceptable. In any case it will always be tied to whether the transformer implements how to transform feature names. So they will never be entirely free for someone implementing an estimator (unless you automatically generate them based on some inherited property).

slep014/proposal.rst
```diff
@@ -0,0 +1,279 @@
...
+For options 2 and 3 the default value of configuration flag:
+``store_feature_names_in`` is False.
+
+Considerations
+##############
+
+Index alignment
+---------------
+
+Operations are index aligned when working with DataFrames. Internally,
```
NicolasHug: It'd be useful to briefly describe what index-alignment means for dataframes, or to provide a link.

For now I share @TomAugspurger opinion and I honestly don't see where index-alignment would ever be a concern for us. So maybe we could just remove the section.

slep014/proposal.rst
```diff
@@ -0,0 +1,279 @@
...
+``pandas``. This leads to unnecessary overhead from piping the data from one
+estimator to another.
+
+Sparse matrices
+---------------
+
+Traditionally, ``scikit-learn`` prefers to process sparse matrices in
+the compressed sparse row (CSR) matrix format. The `sparse data structure <https://pandas.pydata.org/pandas-docs/stable/user_guide/sparse.html>`_ in pandas 1.0, only supports converting directly to
+the cooridnate format (COO). Although this format was designed to quickly
+convert to CSR or CSC formats, the converation process still needs to allocate
```
NicolasHug:
```suggestion
the compressed sparse row (CSR) matrix format. The `sparse data structure <https://pandas.pydata.org/pandas-docs/stable/user_guide/sparse.html>`_ in pandas 1.0 only supports converting directly to
the coordinate format (COO). Although this format was designed to quickly
convert to CSR or CSC formats, the conversion process still needs to allocate
```

slep014/proposal.rst
```diff
@@ -0,0 +1,279 @@
...
+~~~~
+
+- Can be used for heterogeneous data.
+
+Cons
+~~~~
+
+- ``scikit-learn`` does not require many of the features Dataset provides.
+- Needs to be converted to a DataArray before it can be converted to a numpy array.
+- The `conversation from a pandas DataFrame to a Dataset <http://xarray.pydata.org/en/stable/pandas.html>`_
```
NicolasHug:
```suggestion
- The `conversion from a pandas DataFrame to a Dataset <http://xarray.pydata.org/en/stable/pandas.html>`_ 
```

slep014/proposal.rst
```diff
@@ -0,0 +1,279 @@
...
+- Can be used for heterogeneous data.
+
+Cons
+~~~~
+
+- ``scikit-learn`` does not require many of the features Dataset provides.
+- Needs to be converted to a DataArray before it can be converted to a numpy array.
+- The `conversation from a pandas DataFrame to a Dataset <http://xarray.pydata.org/en/stable/pandas.html>`_ 
+  is not lossless. For example, categorical dtypes in a pandas dataframe will
+  lose its categorical information when converted to a Dataset.
```
NicolasHug:
```suggestion
  lose their categorical information when converted to a Dataset.
```

slep014/proposal.rst
```diff
@@ -0,0 +1,279 @@
...
+XArray DataArray
+----------------
+
+`xarray's DataArray <http://xarray.pydata.org/en/stable/data-structures.html#dataarray>`_
+is a data structure that store homogenous data.
+
+Pros
+~~~~
+
+- xarray guartantees that there will be no copies during round-trips from
```
NicolasHug:
```suggestion
- xarray guarantees that there will be no copies during round-trips from
```

slep014/proposal.rst
```diff
@@ -0,0 +1,279 @@
...
+  numpy. (`xarray #3077 <https://github.com/pydata/xarray/issues/3077>`_)
+
+Cons
+~~~~
+
+- Can only be used for homogenous data.
+- As with XArray's Dataset, DataArray does not have much adoption as pandas,
```
NicolasHug:
```suggestion
- As with XArray's Dataset, DataArray does not have as much adoption as pandas,
```

slep014/proposal.rst
```diff
@@ -0,0 +1,264 @@
...
+                 .get_feature_names(orig_cat_cols))
+
+    feature_names = np.r_[cat_names, orig_num_cols]
+
+The ``feature_names`` extracted above corresponds to the features directly
+passed into ``LogisticRegression``. As demonstrated above, the process of
+extracting ``feature_names`` requires knowing the order of the selected
+categories in the ``ColumnTransformer``. Furthemore, if there is feature
+selection in the pipeline, such as ``SelectKBest``, the ``get_support`` method
+would need to be used to select column names that were selected.
```
jnothman:
```suggestion
would need to be used to determine column names that were selected.
```

slep014/proposal.rst
```diff
@@ -0,0 +1,264 @@
...
+default, ``pandas_in_out`` is set to ``False``, resulting in the output of all
+estimators to be a ndarray.
+
+Enabling Functionality
+######################
+
+The following enhancements are **not** a part of this SLEP. These features are
+made possible if this SLEP gets accepted.
+
+1. Allows estimators to treat columns differently based on name or dtype. For
```
jnothman: In what way is this enabled by the present SLEP? I assume this means something more expansive: that we will try to retain dtype in outputting a dataframe e.g. after feature selection.

Otherwise this pertains to the handling of Pandas input, which is done on a case-by-case basis already?

slep014/proposal.rst
```diff
@@ -0,0 +1,264 @@
...
+
+This SLEP proposes attaching feature names to the output of ``transform``. In
+the above example, ``pipe[:-1].transform(X)`` propagates the feature names
+through the multiple transformers.
+
+This feature is only available through a soft dependency on pandas. Furthermore,
+it will be opt-in with the the configuration flag: ``pandas_in_out``. By
+default, ``pandas_in_out`` is set to ``False``, resulting in the output of all
+estimators to be a ndarray.
+
```
jnothman: It's not clear here whether pandas output will be provided even when the input is ndarray.

I don't see why that should only be a matter of "community adoption".

slep014/proposal.rst
```diff
@@ -0,0 +1,264 @@
...
+- **Enforce**: This increases the maintaince burden of third party libraries.
+  This burden includes: checking for the configuration flag, generating feature names and including pandas as a dependency to their library.
+
+- **Not Enforce**: Currently, third party transformers can return a DataFrame
+  or a numpy and this is mostly compatible with ``scikit-learn``. Users with
+  third party transformers would not be able to access the features enabled
+  by this SLEP.
+
+
```
jnothman: Also: code relying on the output of Scikit-learn transformers will presume it is an ndarray presently, but will get something different now.

slep014/proposal.rst
```diff
@@ -0,0 +1,264 @@
...
+Sparse matrices
+---------------
+
+Traditionally, ``scikit-learn`` prefers to process sparse matrices in
+the compressed sparse row (CSR) matrix format. The `sparse data structure <https://pandas.pydata.org/pandas-docs/stable/user_guide/sparse.html>`_ in pandas 1.0 only supports converting directly to
+the coordinate format (COO). Although this format was designed to quickly
+convert to CSR or CSC formats, the conversion process still needs to allocate
+more memory to store. This can be an issue with transformers such as the
+``OneHotEncoder.transform`` which has been optimized to construct a CSR matrix.
+
```
jnothman: Estimators without rectangular input: should they produce pandas output, e.g. how do you get a vectorizer to produce pandas out?

adrinjalali:
>  So @adrinjalali's proposal is ...

Not really. My proposal is that the output only depends on the global flag the user sets to whether or not feature names are enabled. If yes, always return a DataArray (or a DataSet), if not, always return an ndarray, independent of the input.

slep014/proposal.rst
```diff
@@ -0,0 +1,264 @@
...
+estimator to another.
+
+Sparse matrices
+---------------
+
+Traditionally, ``scikit-learn`` prefers to process sparse matrices in
+the compressed sparse row (CSR) matrix format. The `sparse data structure <https://pandas.pydata.org/pandas-docs/stable/user_guide/sparse.html>`_ in pandas 1.0 only supports converting directly to
+the coordinate format (COO). Although this format was designed to quickly
+convert to CSR or CSC formats, the conversion process still needs to allocate
+more memory to store. This can be an issue with transformers such as the
```
rth: This seems to be a fairly conservative statement. Last time I tried to create pandas sparse frame  with 100k features, a year or so ago, that was very slow. We should check whether it's realistic to use pandas sparse format for current scikit-learn use cases.

slep014/proposal.rst
```diff
@@ -0,0 +1,264 @@
...
+XArray Dataset
+--------------
+
+`xarray's Dataset <http://xarray.pydata.org/en/stable/data-structures.html#dataset>`_
+is a multi-dimenstional version of panda's DataFrame.
+
+Pros
+~~~~
+
+- Can be used for heterogeneous data.
```
rth: I would say that handling of sparse is a disadvantage in both,
- xarray supports https://github.com/pydata/sparse as far as I understand, but whether we want to use it and what is the conversion cost remains to be evaluated
- conversion back and forth to pandas sparse frame also has non negligible costs.

amueller:
@rth we're having a meeting discussing the SLEP right now :)
https://www.google.com/url?q=https://anaconda.zoom.us/j/158272910?pwd%3DNURzbWhTSXNYNFhRMVYxRFVybjUrdz09&sa=D&usd=2&usg=AOvVaw0d6Fs7-D2t5Pyz28va20Wd

amueller:
@adrinjalali all of them are conditional on a flag, so I would summarize your proposal as 

Transform Input | Transform Output
--------------- | ----------------
ndarray         | DataArray
DataArray       | DataArray
DataFrame       | DataArray
AnythingElse    | DataArray

rth:
Thanks for working on this it's certainly a challenging topic!

To add to the xarray vs pandas discussion on the subject of sparse arrays; below are quick benchmark results for an input CSR matrix with n_samples=1000, n_features=100k, sparse density=0.01 (i.e. 1M non zero elements):
 - scipy: copy (1ms), CSR->COO (3ms), CSR->CSC (11 ms)
 - pandas: using `pd.DataFrame.sparse.from_spmatrix` (17s **i.e. >2000x slower than CSR->COO conversion**). While things are better in low dimensional space (e.g. with OneHotEncoder) for any application where feature dimensionality is large (e.g. text vectorization) this is IMO not acceptable.  In a different regime, with n_samples=1M, n_features=100, this conversion is ~5x slower than CSR>COO conversion @ 40 ms). pandas stores each column as a `pd.array.SparseArray` as far as I understand (i.e. will create `n_features` such objects) which is not fully equivalent to using CSC, and the reduced performance in high dimensionality seems to be expected https://github.com/pandas-dev/pandas/issues/32196 (no sure if there is a way to address it).

 - xarray: doesn't have a built-in sparse array support, however it works with objects that expose ndarray interface, and in particular [`pydata/sparse`](https://github.com/pydata/sparse). It only supports COO but conversion times with `xarray.DataArray(sparse.COO.from_scipy_sparse(X))` are reasonable (20ms with the original data). It would have also worked with scipy.sparse directly if someone wrote a sparse ndarray class (https://github.com/scipy/scipy/issues/8162).  As a thin wrapper around the data with added labels I do like the xarray approach. Pure speculation, but I imagine it there was an ndarray object that supported categorical dtype it shouldn't be overly complex to make it work with ndarray. The downside of xarray for sparse (at least for now) that it needs to install `pydata/sparse`  which in turns depends on `numba` and `llvmlight`. It may have some performance benefits in the long term, but it's still 3 extra dependencies.


Benchmark code can be found [here](https://github.com/rth/notebooks/blob/master/sklearn/sparse-benchmark.ipynb).

TomAugspurger:
FYI, pandas has our monthly dev call today at 18:00 UTC (~3 hours from this post). I've added some items that came up here to the agenda: https://docs.google.com/document/d/1tGbTiYORHiSPgVMXawiweGJlBw5dOkVJLY-licoBmBU/edit?usp=sharing. All are welcome to join if anyone has anything to add.

GaelVaroquaux:
Another FYI: I've kickstarted a discussion on a dataframe protocol, which
I find very insightful in terms of the discussion that we are having
here: https://discuss.ossdata.org/t/a-dataframe-protocol-for-the-pydata-ecosystem/267/30

If the discussion above would pan-out, it would help in general for the
input, but also for the output (giving us more freedom on the output
choice).

@TomAugspurger memory copies really bother me. Memory usage is a weak
point of our ecosystem. Going out-of-core solves it, but brings in a very
large complexity. Is having 2 data structures in Pandas, with 2 different
memory layout, one optimized for the relational algebra, which would
induce memory copies, one optimized for "primal-space analytics", ie
row-wise addressing, which would not induce memory copies? This tension
with two usecases, and the relating computing tradeoffs in the data
structures is a classic. It is present for instance with the sparse
matrices, for which there are different layout, optimal for different
types of algorithms.

amueller:
I'd really like to have a strategy for moving forward with this. I don't think the outcome of the dataframe discussion is a prerequisite for that.

@adrinjalali suggested that @thomasjpfan writes a proposal / implementation with xarray.

I'm not sure how helpful that will be. It think we need to either agree on a list of evaluation criteria for a proposal or at least some requirements, and a way to actually make a decision. As mentioned on the call, I'm not super in favor of forcing a vote, but also I don't want to keep discussing and making prototypes for ever. There is probably no simple answer. Also, if we vote on individual sleps (like this one) we can also end up in the situation that we reject all proposals, which also doesn't help things.

@GaelVaroquaux re memory copies: sklearn already makes tons of copies, so honestly it's not what I'm most concerned about. If this was a major concern for you, I would have expected this to show up on the roadmap somewhere. We don't even have a process to measure how many copies are being made, right? Also, if it's an opt-in procedure it might not impact existing workflows as much.

Given a standard pipeline with scaling, imputation and OneHotEncoder, how much memory do we need right now? 1x data size? 2x data size? 10x data size? I honestly don't know the answer (though I would be surprised if it's as small as 1 or 2).

GaelVaroquaux:
> @GaelVaroquaux re memory copies: sklearn already makes tons of copies, so honestly it's not what I'm most concerned about. If this was a major concern for you, I would have expected this to show up on the roadmap somewhere.

The French team has been working on reducing them, and on support for float32 over the last year.

Also, there has been a lot of work in joblib / loky to memmap and try to share as much as possible across worker.

So I think that yes, my comment is consistent with our strategy. These efforts amount to several person-years of work.

> We don't even have a process to measure how many copies are being made, right?

memprofiler, and integration in sphinx-gallery (both also pushed forward by us).

> Given a standard pipeline with scaling, imputation and OneHotEncoder, how much memory do we need right now? 1x data size? 2x data size? 10x data size?

Well, the more I study missing values, the more I am convinced that any sophisticated imputation shouldn't be part of a "standard pipeline" :).

I would say, typically x3. I'd like to get this down, not up.

amueller:
I know your team has worked on some cool float32 stuff and I don't want to belittle that work at all. However I haven't seen any benchmarks for a pipeline; and maybe you've done them but not share them?

And I agree, we probably don't want fancy imputation, but we probably still need mean imputation.

Generally I'd love to get the memory copies down, but I don't like making an argument using something for which I haven't seen measurements. I simply don't know if there is a common case where this copy would make or break a workflow.

GaelVaroquaux:
Fair enough. We should prototype and measures before doing any choice. I agree with you.

amueller:
Anyway, my main point is that we should make a plan on how we will make a
decision.

amueller:
So what would you want to prototype and measure and how would you make
decisions based on the outcome?

I'm pretty sure any outcomes would depend on the choice of pipeline that's
used and I can imagine that under some conditions pandas-in pandas-out will
require less memory, while it will probably require slightly more in most
cases (assuming a memory copy there which we can't really benchmark because
that implementation of pandas doesn't exist).

I'm ok with setting up goals and criteria but I think just writing up one
more prototype (we have like 5 or so right now?) will help us make a
decision.

Is there an amount of memory overhead we're willing to accept? And how
would we trade memory overhead vs user interface cost?

jnothman:
I also think that worrying about a memory copy that might happen in the
future only if pandas implements things a certain way, only if the user
provides pandas input (while ndarray input continues to be cheap) and while
the status quo and many valuable DataFrame use cases already entail a copy,
is probably premature optimisation. We should likely be biased towards
pandas because that's what users already expect to work.

TomAugspurger:
I share Joel's outlook: the potential pandas refactor that would induce the additional memory copy is uncertain. It's some years off at a minimum, and I wouldn't be surprised if it never happened. We discussed this at our dev call on Wednesday (which @thomasjpfan joined) but aren't at the point where we can give any guidance; things are just too uncertain.

---

Not to derail things with a new proposal at this late hour, but... we have two potential contenders for <labeled array in> -> <labeled array out>. Which means a protocol! Scikit-learn could define something like (With dummy implementations for DataFrame).

```
def __sklearn_feature_names__(self) -> List[str]:
    return list(self.columns)


def __sklearn_data__(self, ...):  # dtype? Anything else?
    """Return the values for the estimator."""
    return self.to_numpy()

def __sklearn_transformed_result__(self, original, transformed, feature_names):  # original?
    """Wrap the result"""
    return pd.DataFrame(transformed, index=original.index, columns=feature_names)
```

I'm sure pandas (and likely xarray) would happily implement that protocol. (We're also more comfortable diving into our internals to avoid memory copies, but that's another story). However, I think it's premature to set down a protocol. I personally don't have a good opinion on what sorts of arguments would be helpful in these methods. I think we might want to learn a bit first, before freezing the API in a protocol.

NicolasHug:
> So what would you want to prototype and measure

We set up a bunch of realistic pipelines and check the memory usage with memory_profiler. We can also set up fake pipelines with dummy transformers that never/always copy for a best/worst case scenario. We can simulate pandas potential future behavior by artificially introducing copies, e.g. with a wrapper for dataframes.


-----

Now about copies, dumb question: how bad can it be?

Say we pass `X_orig` to a pipeline of transformers that, for the sake of the argument, makes copies of the input at every single step.

`X_orig -> STEP1 (copy) -> ... STEPN (copy)`

Each transformer looks like

```py

def fit_transform(self, X):
	X_copy = validate(X)  # make a copy
	X_out = # allocate output
	# do the work ... 
	return X_out
```

So at any point during the execution, we have *at most* 4 datasets in memory:

- `X_orig`
- the input `X`
- the copy `X_copy`
- the output `X_out`.

In a best case scenario where we never copy, we still have *at least* 3 datasets (we just don't have `X_copy`).

That's only 1 extra dataset for the worst case scenario then?

TomAugspurger:
I had a realization last night (which I think was already known to others like @jreback).
I think that *the memory copy concern may not be an issue even if / when pandas goes to a
column store in the future.* The rest of this post is going to go into some internal
pandas details.

Consider a user with a DataFrame `df` and a pipeline `make_pipeline(StandardScaler(), PCA())`

There's two places where we need to watch out for memory copies:

1. DataFrame -> ndarray (inside `check_array` say)
2. 2D ndarray -> DataFrame (inside the DataFrame constructor)

Let's actually start with the second case, which is pretty easy and sets up the first. In a column-store future,
pandas will "split" this 2D ndarray into a sequence of 1D arrays. *Each of these 1D arrays will be views on the
original memory*, so no memory copies.

```python
In [2]: a = np.ones((10, 5))

In [3]: slices = {f'{i}': a[:, i] for i in range(a.shape[1])}

In [4]: df = pd.DataFrame(slices)
In [5]: df._data.blocks  # 2D array split into 5 "blocks". This doesn't happen today, but may in the future
Out[5]:
(FloatBlock: slice(0, 1, 1), 1 x 10, dtype: float64,
 FloatBlock: slice(1, 2, 1), 1 x 10, dtype: float64,
 FloatBlock: slice(2, 3, 1), 1 x 10, dtype: float64,
 FloatBlock: slice(3, 4, 1), 1 x 10, dtype: float64,
 FloatBlock: slice(4, 5, 1), 1 x 10, dtype: float64)

In [6]: df._data.blocks[0].values.base  # each block is a view on 'a'
Out[6]:
array([[1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.]])
```

(@jorisvandenbossche can confirm that you can create pyarrow arrays as a view on
and ndarray with no missing values for primative types like int64?)

Now, for potential copy 1, which is where pandas can get creative in `DataFrame.__array__`.
For the special case of `np.asarray(DataFrame(2d_array))`, we an check that the `.values.base`
of each array is the same object. In this case, we can just return that `.base`, no memory copy!

```python
In [19]: b0 = df._data.blocks[0].values.base

In [20]: if all(blk.values.base is b0 for blk in df._data.blocks):
    ...:     arr = b0
    ...:

In [21]: arr
Out[21]:
array([[1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.]])

```

So tldr: I think we're able to do things without memory copies from 2D ndarray -> DatFrame -> 2d ndarray, even in a future where pandas has only 1D blocks.

jnothman:
There seems to be some consensus here that we should be sticking to pandas out for now. Any strong objections? Can we progress on resolving the comments and moving towards vote?

adrinjalali:
+1

thomasjpfan:
Okay I will move this forward.

The only thing I am concerned about is the sparse performance (as noted in https://github.com/scikit-learn/enhancement_proposals/pull/37#issuecomment-596577388).

Should we hedge a little and adjust the configuration flag to:

```py
set_config(array_in_out='pandas')
```

where the default is `'ndarray'`. This leaves open the possibility of supporting other dataframe-like objects.

NicolasHug:
Following up on https://github.com/scikit-learn/enhancement_proposals/pull/37#issuecomment-598449861, I too would be more comfortable if I could inform my voting decision with benchmark results.

amueller:
@NicolasHug what would you measure? @TomAugspurger laid out a zero-copy strategy. So would you want to do a prototype implementation of that and then measure if there's any unforeseen consequences?
Or do you mean with the current implementation that we also expect to have no memory copy to confirm that there is indeed none?

amueller:
@thomasjpfan just to be clear, your proposal is to always create a pandas dataframe for ``array_in_out='pandas'``, even if the input is an ndarray or sparse matrix, right?

I mean we *could* have an option ``'pandas'`` and an option ``'pandas_if_pandas_in'`` but I'm not sure how user-friendly that is.

If we always return pandas, I am a bit concerned in the sparse case, because that will mean a memory copy from my understanding, right?

Btw, @adrinjalali in your proposal, did you also suggest always returning a DataArray for sparse data?

My understanding of @thomasjpfan's proposal is have an option to do

Transform Input | Transform Output
--------------- | ----------------
ndarray         | DataFrame
scipy.sparse | DataFrame
DataArray       | DataFrame
DataFrame       | DataFrame
AnythingElse    | DataFrame

with the potential to in the future have the same for ``DataArray`` by doing ``array_out='xarray'``.

I wonder if 

Transform Input | Transform Output
--------------- | ----------------
ndarray         | ndarray
scipy.sparse | scipy.sparse
AnythingElse    | DataFrame

or even

Transform Input | Transform Output
--------------- | ----------------
ndarray         | DataFrame
scipy.sparse | scipy.sparse
AnythingElse    | DataFrame

might be more effective.

rth:
> I am a bit concerned in the sparse case, because that will mean a memory copy from my understanding, right?

It's much worse than a memory copy for sparse cf https://github.com/scikit-learn/enhancement_proposals/pull/37#issuecomment-596577388 (at least in high dimensionality)

rth:
> I wonder if
> Transform Input | Transform Output
> scipy.sparse | scipy.sparse
> might be more effective.

It could be, but then one couldn't get feature names for a simple pipeline of OneHotEncoder + LinearModel since OHE returns sparse data by default. Which is also likely not ideal.

amueller:
@rth Yes, sorry, it's a conversion + copy. Suggestions? ;)

NicolasHug:
>@NicolasHug what would you measure?

please see https://github.com/scikit-learn/enhancement_proposals/pull/37#issuecomment-598714355

>  @TomAugspurger laid out a zero-copy strategy

Everything is very hypothetical still, and I'd rather base my decision on facts and numbers than on hypothetical behaviors.

I'm not against calling for a vote, but I don't have enough info to make up my mind yet

TomAugspurger:
> Everything is very hypothetical still, and I'd rather base my decision on facts and numbers than on hypothetical behaviors.

Just to clarify, the *behavior today* is exactly what scikit-learn wants. For homogenous, numeric data, `asarray(DataFrame(ndarray))` is zero copy:

```python
In [10]: a = np.random.randn(100, 10)

In [11]: df = pd.DataFrame(a)

In [12]: b = np.asarray(df)

In [13]: b.base is a
Out[13]: True
```

The "hypothetical" bit is going to have to remain hypothetical, since it's uncertain whether that refactor will actually happen.

amueller:
> Everything is very hypothetical still, and I'd rather base my decision on facts and numbers than on hypothetical behaviors.

You suggest evaluating a worst-case bound based on a forced single memory copy. That gets you a worst-case bound for a very specific worst-case. If the worst-case bound is "not that bad" and the scenario is likely that would be indeed helpful. However, the situation for sparse is much worse than your worse-case scenario would imply.  So an "even worse" case would be to have similar overhead for dense, which would be worse than what's covered by your benchmark, because your benchmark also uses a very specific hypothetical future ;)

NicolasHug:
Sorry @amueller I don't completely follow: we can run the benchmarks for both dense and sparse data? I don't see yet how my comment describes a "very specific worst case".

What I'm trying to say is, I'd be happy to vote yes if the benchmarks show that the worst case scenario isn't that bad after all. (as long as we can simulate the actual worst case).

For now, IMHO, there is no tangible element on which to base a decision (hypothetical future behavior of pandas + hypothetical solution for that hypothetical behavior).

I'm basically following your own words:

> I don't like making an argument using something for which I haven't seen measurements

But like I said, if others want to vote, let's vote.

amueller:
@NicolasHug

We can certainly run a benchmark for what's there currently, but we also know the outcome: for dense data there is no overhead, and for sparse data there is unacceptably large overhead.

What you also proposed is a hypothetical future scenario in which there is a single copy in the case of dense data (and an additional copy in the case of sparse data?). What I tried to say is that at least theoretically the future could be even worse than that, and so this benchmark also makes assumptions about the future implementation.

amueller:
I don't know what the current proposal is for sparse data so I would like to clarify that first, given the issues that @rth pointed out above.

jorisvandenbossche:
> [@GaelVaroquaux] memory copies really bother me. Memory usage is a weak
point of our ecosystem. .. Is having 2 data structures in Pandas, with 2 different
memory layout, one optimized for the relational algebra, which would
induce memory copies, one optimized for "primal-space analytics", ie
row-wise addressing, which would not induce memory copies? 

Memory copies are also a big problem in pandas, and actually one of the reasons to go towards a column store in pandas (to avoid things like "consolidation" of same-dtype columns etc). 

So while I personally am pushing for a column-store in pandas, I still think there is value in having a kind of homogeneously typed DataFrame backed by a single 2D array (we talked about this briefly in the call last week). So it would be a worthwile discussion to think about where in the PyData ecosystem that fits. Should people just use xarray for that use case? Or can we have such a single dtype DataFrame class in pandas that can share some infrastructure with the column-store DataFrame? That are interesting discussions, but let's have that discussion somewhere else ;)

But, as @amueller / @jnothman already argued, this is not really a concern right now, and as @TomAugspurger showed, it seems even possible to hack together a zero-copy roundtrip even if DataFrame would become a column-store.

> (@jorisvandenbossche can confirm that you can create pyarrow arrays as a view on
and ndarray with no missing values for primative types like int64?)

Yes, that is possible (even for floats with NaNs that is possible, if you keep them as NaNs and don't convert to missing vlaues).


> > I am a bit concerned in the sparse case, because that will mean a memory copy from my understanding, right?
>
> It's much worse than a memory copy for sparse cf #37 (comment) (at least in high dimensionality)

@rth I just commented in https://github.com/pandas-dev/pandas/issues/32196 that this can be sped up rather easily. But of course, it will always stay a lot slower since it is creating many 1D sparse arrays from the 2D array. 
In principle, I think this can even be zero-copy as well (although it will still be slow).

rth:
> I just commented in pandas-dev/pandas#32196 that this can be sped up rather easily.

Cool, thanks @jorisvandenbossche ! That would indeed resolve most of the concerns with sparse.

jorisvandenbossche:
I am not sure that resolves most of the concerns regarding sparse. First, it's only possibly in a future release. And moreover, even if we can give that conversion a factor 10 speed-up in pandas, it is still *much* slower than a CSR->CSC conversion.

For sparse output of transformers, it seems still best to output scipy matrices, but not sure how to deal with features names then.

rth:
Well as far as I know, there is no ideal good way currently to have sparse arrays with labels in general. The alternative with xarray is to use [`pydata/sparse`](https://github.com/pydata/sparse). That requires numba/llvmlite (i.e. we need to add optional dependencies), and there are some performance concerns for basic operations https://github.com/pydata/sparse/issues/331 if users actually want to do anything with the return objects. 

So given that this is going to be an opt-in feature and only available in ~7 months in 0.24, as far as I understand, saying that you need latest pandas for best performance with sparse by that time could be reasonable IMO. I think we might want to compare  to the cost on a typical workflow rather than to CSR->CSC conversion. Currently the performance blocker is only with a very large number of features (e.g. 100k). As mentioned in previous benchmark for `n_samples=1M, n_features=100`, it is 5x slower than CSR->CSC  conversion but in the end only 200ms so likely not that noticeable with respect to the overall pipeline cost (particularly if we can make the conversion 10X faster).

rth:
> What I'm trying to say is, I'd be happy to vote yes if the benchmarks show that the worst case scenario isn't that bad after all. 

This discussion shows that we currently don't have benchmarks for a few typical pipelines, and cannot easily evaluate the performance impact of proposed changes.  It would be good to merge ASV benchmarks created by @jeremiedbb into scikit-learn (https://github.com/scikit-learn/scikit-learn/issues/16723), use them as a reference and extend them as needed. I really like how this is [done and documented in pandas](https://pandas.pydata.org/docs/development/contributing.html?highlight=asv#running-the-performance-test-suite). Not a blocker for this SLEP but would be nice to have.

amueller:
I think we should ask ourselves whether we're ok with having a solution that doesn't address sparse. It would certainly mean not covering some important use-cases.
The other option is to go back to having our own datatype. Or our own datatype for sparse. Or using xarray+pydata/sparse. hm...

TomAugspurger:
> we could see if pandas could at least improve on the conversions maybe the same way that it's done in pydata/sparse

@jorisvandenbossche and @rth have been doing some work on pandas to get converting between a scipy.sparse matrix and pandas' format faster (hopefully they can report results). But even with this improvements we're still going to bump against the fact that pandas' SparseArray is 1-D, so when you have a sparse results with ~10,000s of columns, you're creating ~10,000s of Python objects, which is slow, even if creating each individual object is cheap. pandas' extension arrays being 1-D is a fundamental limitation that isn't likely to change soon.

So there does seem to be a choice between "pandas in / pandas out" and "fast sparse transformed results" which may warrant its own configuration.

thomasjpfan:
I am currently prototyping this feature out, so we can measure what the overhead looks like for the various configurations combinations of xarray/pandas/sparse. I am planning to get this out this week.

amueller:
@TomAugspurger I'm a bit hesitant to have two separate configuration options to have some form of feature names. Or are you saying if we enable pandas output we would also go to pandas for sparse, and then have another option to move to a different format?
And what would that look like?

TomAugspurger:
I just worry that because pandas' support for very wide sparse datasets is so poor, people will be surprised / annoyed at `set_config(pandas_in_out=True)` slowing down things returning spare results. I'm sure there's a bit more we can do on the pandas side to optimize constructing a DataFrame from a sparse object, but it'll always have substantial overhead over scipy as long as we're creating one 1-D array per column.

rth:
> it'll always have substantial overhead over scipy as long as we're creating one 1-D array per column.

Let's way for benchmark results from https://github.com/scikit-learn/scikit-learn/pull/16772 to decide. The question is not so much the performance difference between pandas sparse and scipy sparse for constructors, but how much this would affect real word pipelines. <10% overhead to get feature names might be OK, x10 in some use cases is definitely not.

> people will be surprised / annoyed at set_config(pandas_in_out=True) slowing down things returning spare results. 

I think returning some custom object or a pydata/sparse wrapped into an xarray would results in as much suprise / annoyance for users not familiar with either.

amueller:
One thing that I didn't realize before [maybe it was discussed above and I just forgot?] is that deciding output type based on input type is not possible for CountVectorizer, which will be one of the main use-cases. @thomasjpfan implemented the 'we always output <target data structure>' option, which produces array/dataframe/xarray depending on the configuration setting, independent of the input.

jnothman:
What are the next steps here?

thomasjpfan:
Given the developments in https://github.com/scikit-learn/scikit-learn/pull/16772, this would amount to a new SLEP which describes `array_out`, which is different from "pandas in pandas out". It should describe the limits that were discovered in https://github.com/scikit-learn/scikit-learn/pull/16772, i.e. sparse matrices are better with xarray, etc.

amueller:
This should be moved to rejected as we accepted SLEP 18, right?

thomasjpfan:
In [`653e476` (#37)](https://github.com/scikit-learn/enhancement_proposals/pull/37/commits/653e476588537b735f973f80bf211e32f96920da), this SLEP is now rejected.

PR #38, opened 2020-02-19
make more explicit that verbose_feature_names is not required

amueller:
Addressing my concern in #32, this rephrases the ``verbose_feature_names`` to be an optional extension.

### DIFF
--- a/slep007/proposal.rst
+++ b/slep007/proposal.rst
@@ -111,19 +111,6 @@ original features:
 This proposal talks about how feature names are generated and not how they are
 propagated.
 
-verbose_feature_names
-*********************
-
-``verbose_feature_names`` controls the verbosity of the generated feature names
-and it can be ``True`` or ``False``. Alternative solutions could include:
-
-- an integer: fine tuning the verbosity of the generated feature names.
-- a ``callable`` which would give further flexibility to the user to generate
-  user defined feature names.
-
-These alternatives may be discussed and implemented in the future if deemed
-necessary.
-
 Scope
 #####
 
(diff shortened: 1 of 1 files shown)


slep007/proposal.rst
```diff
@@ -267,7 +254,29 @@ and hence the relevance of ``verbose_feature_names=False``::
...
+
+- an integer: fine tuning the verbosity of the generated feature names.
+- a ``callable`` which would give further flexibility to the user to generate
+  user defined feature names.
+
+These alternatives may be discussed and implemented in the future if deemed
+necessary.
+
+In case of the ``ColumnTransformer`` example above ``verbose_feature_names``
+could remove the estimator names, leading to shorter and less redundant names::
```
adrinjalali: This is contradicting the `By default, transformers will retain existing feature names.` statement.

I guess the default would not be the same for all transformers. WDYT?

slep007/proposal.rst
```diff
@@ -267,7 +254,29 @@ and hence the relevance of ``verbose_feature_names=False``::
                          'ohe_make_ABC', 'ohe_make_XYZ', ...,
                          'pca_pca0', 'pca_pca1', 'pca_pca2']
 
-If desired, the user can remove the prefixes::
+Extensions
+##########
+
+verbose_feature_names
+*********************
+By default, transformers will retain existing feature names. In some cases
```
adrinjalali: we could probably just saye "most transformers" here to fix the inconsistency issue.

slep007/proposal.rst
amueller: I agree it's contradictory. But it's actually also contradictory to what's in the Feature Generating Transformers section. I think if we want to have ``verbose_feature_names`` as an optional extension then this section also needs to be rewritten.

amueller:
I moved all mentions of ``verbose_feature_names`` to extensions and defined the behavior without it for all cases. I think it's important that the SLEP contains enough information so it can actually be implemented without a lot of further discussion.

adrinjalali:
@NicolasHug WDYT?

slep007/proposal.rst
```diff
@@ -172,17 +157,12 @@ Feature Generating Transformers
...
-...``. An alternative is to leave the feature names unchanged when each output
-feature corresponds to exactly one input feature. Whether or not to modify the
-feature name, *e.g.* ``log(x0)`` vs. ``x0`` may be controlled via the
-``verbose_feature_names`` to the constructor. The default value of
-``verbose_feature_names`` can be different depending on the transformer. For
-instance, ``StandardScaler`` can have it as ``False``, whereas
-``LogTransformer`` could have it as ``True`` by default.
+generate a column based on a single given column. These would simply
+preserve the input feature names if a single new feature is generated,
+such as in ``StandardScaler``, which would mape ``'age'`` to ``'age'``.
```
NicolasHug: map
```suggestion
such as in ``StandardScaler``, which would map ``'age'`` to ``'age'``.
```

adrinjalali:
> I think these PRs could be merged with only 1 approval, especially if the approval comes from the "SLEP champion".

I agree, but I still wanted to know what you think first :)

PR #61, opened 2021-11-28
ENH Adds the details about the numpy arrays and strings

thomasjpfan:
Adds more details to SLEP007 regarding dtypes, containers and strings.

This PR is a follow up to https://github.com/scikit-learn/enhancement_proposals/pull/59#issuecomment-962448084

CC @glemaitre @adrinjalali @jnothman

### DIFF
--- a/slep007/proposal.rst
+++ b/slep007/proposal.rst
@@ -129,17 +129,19 @@ Input Feature Names
 The input feature names are stored in a fitted estimator in a
 ``feature_names_in_`` attribute, and are taken from the given input data, for
 instance a ``pandas`` data frame. This attribute will be ``None`` if the input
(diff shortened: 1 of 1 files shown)


adrinjalali (approved):

jjerphan (approved):

glemaitre (approved):
LGTM for the changes.

I just realised that the `post-history` in PEP is only related to the mailing list posting to keep up-to-date with the community (something that we could think about in the future).

There is no date to acknowledge any change in a PEP. Basically, changes can be done to reflect the implementation but without any extra layer of complexity. 

Since these changes here are not controversial, I will merge them. We can discuss later the SLEP process and if we want to have a more complex one.

PR #69, opened 2022-06-07
Add X,y,Y to ignored list

adrinjalali:
`X,y,Y` are automatically ignored in the routing mechanism when we infer routed parameters.

xref: https://github.com/scikit-learn/scikit-learn/pull/23342#discussion_r891331438

cc @thomasjpfan

### DIFF
--- a/slep006/proposal.rst
+++ b/slep006/proposal.rst
@@ -77,8 +77,8 @@ the time of writing this proposal, all metadata requested in the core library
 are sample aligned.
 
 Also note that ``X``, ``y``, and ``Y`` input arguments are never automatically
-added to the routing mechanism and they are routed and passed around the way
-they have been passed around until now.
+added to the routing mechanism and are always passed into their respective
+methods.
 
 Detailed description
 --------------------


slep006/proposal.rst
```diff
@@ -76,6 +76,10 @@ Note that in the core library nothing is requested by default, except
 the time of writing this proposal, all metadata requested in the core library
 are sample aligned.
 
+Also note that ``X``, ``y``, and ``Y`` input arguments are never automatically
+added to the routing mechanism and they are routed and passed around the way
+they have been passed around until now.
```
thomasjpfan, line 81: For those that do not know what the previous behavior:

```suggestion
added to the routing mechanism and are always passed into their
respective methods.
```

amueller (approved):

jjerphan (commented):
Thanks for adding this mention, @adrinjalali.

Based on [the specification of SLEPs' "Review and Resolution" defined by SLEP000](https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep000/proposal.html#review-and-resolution), shouldn't we change the status of SLEP006 to `Provisional` (or `Provisionally Accepted`) as it is still being slightly adapted?

adrinjalali:
I think it's not the SLEP which is changing, they're implementation details. I don't think it'd be healthy for this SLEP to go back to not being accepted.

jjerphan:
Yep, from my understanding of SLEP000, I not think `Provisional` means it is not `Accepted`. But nevermind: this is just a small interjection.

jjerphan (approved):

slep006/proposal.rst
```diff
@@ -76,6 +76,10 @@ Note that in the core library nothing is requested by default, except
 the time of writing this proposal, all metadata requested in the core library
 are sample aligned.
 
+Also note that ``X``, ``y``, and ``Y`` input arguments are never automatically
```
thomasjpfan, line 79: This is the only mention of upper-case `Y` in this document. (I suspect it is for `PLS`) Do you think this will be confusing?

adrinjalali: I mean, from the perspective of the SLEP and the implementation, we don't really care why it's here. These arguments never enter the routing scheme.

jjerphan: Small interjection to bring information: `Y` is also used as an argument for some [`pairwise_distances`](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html)-based APIs.

thomasjpfan: In case of `pairwise_distance`, `Y` is a feature array. In the context of this SLEP, `X` and `y` are feature and target data. I am okay with leaving this statement as is.

thomasjpfan (approved):

jnothman (approved):

adrinjalali:
Then I'll merge this with 4 approvals.

PR #81, opened 2022-11-20
SLEP019: Governance Update - Recognizing Contributions Beyond Code (version II)

cmarmo:
Dear all,
this pull request offers modifications to the SLEP019 draft, following the discussion in  #74.
The rendering is available [here](https://scikit-learn-enhancement-proposals--81.org.readthedocs.build/en/81/slep019/proposal.html).

PR #82, opened 2022-11-30
slep019: Add non-core-dev contributors to an ack section

adrinjalali:
I think something which we haven't done so far, but would be nice to do, is to acknowledge people who contribute to the development of a slep, but are not core contributors.

I'm not sure of the text, or whom to include/exclude, but I thought I could open a PR and see how y'all feel about it. I've included Reshama and Lucy as well since at the time of writing the slep, they're not a part of the to-be-formed "core contributor" team.

I haven't added @cmarmo here since she's an author with #81 on the slep.

So, WDYT?

### DIFF
--- a/slep019/proposal.rst
+++ b/slep019/proposal.rst
@@ -144,7 +144,7 @@ will replace a vote on the private Core Contributors' mailing list.
 Acknowledgment
 **************
 
-We talk the following people who have helped with discussions during the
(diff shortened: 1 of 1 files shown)


jjerphan (approved):
I would just target the branch of #81.

jnothman (approved):

PR #86, opened 2023-03-09
SLEP021: Unified API for computing feature importance

glemaitre:
Reflection around a unified API to compute feature importance in scikit-learn

### DIFF
--- a/slep021/proposal.rst
+++ b/slep021/proposal.rst
@@ -1,0 +1,36 @@
+.. _slep_021:
+
+==================================================
+SLEP021: Unified API to compute feature importance
+==================================================
+
+:Author: Thomas J Fan, Guillaume Lemaitre
+:Status: Draft
+:Type: Standards Track
+:Created: 2023-03-09
+
+Abstract
+--------
+
+Detailed description
+--------------------
+
+Discussion
+----------
+
+References and Footnotes
+------------------------
+
+.. [1] Each SLEP must either be explicitly labeled as placed in the public
+   domain (see this SLEP as an example) or licensed under the `Open Publication
+   License`_.
+.. [2] `scikit-learn Governance and Decision-Making
+   <https://scikit-learn.org/stable/governance.html#decision-making-process>`__
+
+.. _Open Publication License: https://www.opencontent.org/openpub/
+
+
+Copyright
+---------
+
+This document has been placed in the public domain. [1]_


glemaitre:
ping @thomasjpfan. Feel free to edit and push directly in the branch.

adrinjalali (commented):
Another API suggestion would be to introduce a `FeatureImportanceCalculatorInterface` kinda thing and pass that either to `__init__`, or plotting methods, or to `get_feature_importance()`, or the other way around, to pass an estimator to it, which is like the meta-estimator idea you have here.

It's a good start here, but this is gonna be a bumpy road lol

slep021/proposal.rst
```diff
@@ -0,0 +1,192 @@
...
+Detailed description
+--------------------
+
+Motivation
+~~~~~~~~~~
+
+Data scientists rely on feature importance when inspecting a trained model.
+Feature importance is a measure of how much a feature contributes to the
+prediction and thus gives insights on the model the predictions provided by
+the model.
```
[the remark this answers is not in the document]
glemaitre: I am fine keeping it high level without going into details. The only important part that should come afterwards is that one needs to provide some data to compute these importances.

slep021/proposal.rst
```diff
@@ -0,0 +1,192 @@
...
+importance. :func:`sklearn.inspection.permutation_importance` already provides
+a way to do that since it repeats the computation several time. For the model
+specific feature importance, the user can use cross-validation to get an idea
+of the dispersion::
+
+   >>> cv_results = cross_validate(tree, X_train, y_train, return_estimator=True)
+   >>> feature_importances = [est.feature_importances_ for est in cv_results["estimator"]]
+   >>> plt.boxplot(feature_importances, labels=X_train.columns)
+
+The second usage is about the model selection. Meta-estimator such as
```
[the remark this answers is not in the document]
glemaitre: indeed feature selection

jnothman:
Definitely a user trap worth removing!

Are there cases where selecting the feature importance method(s) results in different work/storage of information during `fit`?

slep021/proposal.rst
```diff
@@ -0,0 +1,202 @@
...
+~~~~~~~~
+
+A common API
+^^^^^^^^^^^^
+
+**Proposal 1**: Expose a parameter in `__init__` to select the method to use
+to compute the feature importance. The computation will be done using a method,
+e.g. `get_feature_importance` that could take additional parameters requested
+by the feature importance method. This method could therefore be used
+internally by :class:`sklearn.model_selection.SelectFromModel`.
```
vitaliset, line 139: Maybe I'm violating some SOLID principle, but we could incorporate feature importance agnostic techniques into some mixin like `ClassifierMixin` and `RegressorMixin` and specific methods within each class when applicable. In this sense, we would have the "main" feature importance method selected during __init__ (defining the behavior of `get_feature_importance`). Still, one could always use the others because the estimator would have `get_permutation_importance`, `get_mdi_importance`, `get_abs_coef_importance` etc.

slep021/proposal.rst
```diff
@@ -0,0 +1,202 @@
...
+importance. It seems that they should all at least accept a fitted estimator,
+some dataset, and potentially some extra parameters.
+
+**Proposal 3**: Similarly to the proposal 2 and taking inspiration from the
+SHAP package [2]_, we could create a class `Explainer` providing a
+`get_feature_importance` method given some data.
+
+Currently scikit-learn provides only global feature importance. The previous
+API could be extended by providing a `get_samples_importance` to compute an
+explanation per sample if the given method supports it (e.g. Shapley values).
```
vitaliset, line 158: About "`explain_one`", I'm trying to think of scenarios where you can use it and "`explain_all`" inside the same `Explainer`. I don't see many different scenarios than doing something like `explain_all(X) = np.mean(np.abs(explain_one(X)), axis=1)` (for SHAP or LIME for instance). Did you have any other technique in mind?

slep021/proposal.rst
```diff
@@ -0,0 +1,202 @@
...
+
+**Proposal 4**: Create a meta-estimator `FeatureImportanceCalculator` that
```
vitaliset, line 160: I still need to choose my favorite proposal, although they are all very elegant (superficially thinking, I prefer the functions option). But I find proposal 1 less complex than adding an extra layer of abstraction proposed in 2, 3, and 4 (creating meta-estimators).
From the user's point of view, not having to "leave the estimator class" facilitates and increases the chance of use, at least for a scikit-learn noob. I don't know if this is a concern. I always see the scikit-learn development following this path and wanted to understand why. Is it just an object orientation issue?

vitaliset (commented):
I'm super excited about this proposal! Congratulations, @glemaitre and @thomasjpfan, for structuring this!

From my point of view, we also have the option of it always being a function (just like for `sklearn.inspection.permutation_importance`). Any reason it's not listed? (Sorry if I didn't follow some discussion that discourages this option.)

slep021/proposal.rst
glemaitre: Theoretically speaking, I would be surprised that the average of the absolute would always be the way to go. Having a class would allow us to redefine the proper way to combine individual explanations.

glemaitre (commented):
My main thought regarding the usage of a class upon a function is to be able to store states. Having states can also help for efficient computation with potential caching mechanisms. One solution that I did not mention here is to incorporate the plotting capabilities directly inside the meta-estimator. This would also be new.

slep021/proposal.rst
```diff
@@ -0,0 +1,214 @@
...
+importance:
+
+- The function :func:`sklearn.inspection.permutation_importance`. It requests
+  a fitted estimator and a dataset. Addtional parameters can be provided. The
+  method returns a `Bunch` containing 3 attributes: all decrease in score for
+  all repeatitions, the mean, and the standard deviation across the repeats.
+  This method is therefore estimator agnostic.
+- The linear estimators have a `coef_` attributes once fitted, which is
+  sometimes used their corresponding importance. We documented the limitations
+  when it comes to interpret those coefficients.
```
adrinjalali, line 55: a link to the docs for this would be nice for the record.

slep021/proposal.rst
```diff
@@ -0,0 +1,214 @@
...
+features.
+
+In both cases, it should be better to request the user to be more explicit and
+request to choose a specific method to compute the feature importance for
+inspection or feature selection.
+
+Additionally, `feature_importances_` and `coef_` are statistics derived from
+the training set. We already documented that the reported
+`feature_importances_` will potentially show biases for features used by the
+model to overfit. Thus, it will potentially negatively impact the feature
```
adrinjalali, line 122: also a link to the docs we're mentioning would be nice

adrinjalali (commented):
At least to me, the one proposal which I had liked was to have a method at the estimator level, replacing `feature_importances_` and taking all the arguments relevant to that method for that estimator, which I don't see in the proposals section here. We could also add different methods to explain both samples and features separately maybe?

I also don't mind the `Explainer`, but that might be a tricky one to implement?

Whatever we decide, we should make sure it's easy to add explainer methods w/o too much `__init__` param overhead via separate classes or separate methods.

This proposal should also provide pros and cons between different alternatives and suggest one to move forward. WDYT of having a draft PR for the one/two that we like to have a better idea?

lionelkusch:
I found this representation, which can help you to identify the different interests in basic methods and their applicability: 
analyzing_features

[link](https://www.levelingupwithxai.com/leveling-up-with-xai/xai-guide/analyzing-feature-effects)

PR #88, opened 2023-05-12
SLEP022: Fixing randomness ambiguities

NicolasHug:
Disclaimer: I won't be able to champion this SLEP.
I'm opening it here now because I hope it can help better framing the discussions happening in https://github.com/scikit-learn/scikit-learn/issues/26148.


betatim:
I'd like to take on this SLEP if it is Ok with you.

NicolasHug:
Sure, happy for you to take over @betatim , thanks!

slep022/proposal.rst
```diff
@@ -0,0 +1,703 @@
...
+    Play around with `this notebook
+    <https://nbviewer.jupyter.org/gist/NicolasHug/2db607b01482988fa549eb2c8770f79f>`_
+    which illustrates the proposed solution.
+
+    Please, **please** take some time to read and understand this SLEP
+    before taking part in discussions.
+
+
+Abstract
+========
```
betatim, line 27: Mention that we make these changes in behaviour together with the switch to the new Numpy random infrastructure. Combined with the plan of `random_state=` to `rng=` transition for the argument name.

slep022/proposal.rst
```diff
@@ -0,0 +1,703 @@
...
+`fit`, `split`, and `get_params` are unchanged.
+
+In order to **explicitly** support use-cases C and D, CV procedures like
+`cross_val_score` should be updated with a new `use_exact_clones` parameter::
+
+    def _check_statistical_clone_possible(est):
+      if 'random_state' not in est.get_params():
+          raise ValueError("This estimator isn't random and can only have exact clones")
+
+    def cross_val_score(est, X, y, cv, use_exact_clones=True):
```
betatim, line 193: Rename `use_exact_clones` so that statistical clones are allowed but it is fine to have deterministic estimators as well.

Maybe `allow_statistical_clone=`

slep022/proposal.rst
```diff
@@ -0,0 +1,703 @@
...
+The advantage is that we respect our convention with `__init__`.
+
+However, `fit` idempotency isn't respected anymore: the first call to `fit`
+clearly influences all the other ones. The mental model is also not as clear
+as the one of the proposed solution: users don't really know when that seed
+is going to be drawn unless they know the internals of the procedures they
+are using.
+
+This also introduces a private attribute, so we would need more intrusive
+changes to `set_params`, `get_params`, and `clone`.
```
ogrisel, line 409: I think we could alternatively explore the possibility to do stateless consumption in `fit` instead of seeding in `__init__`.

```python
    def fit(self):  # or split()
        rng = check_random_state(self.random_state, copy=True)
        ...
```

This way we would still have the stateless fit semantics for the `est.random_state` attribute.

I will try to find the time to explore this option an alternative notebook in a gist.

ogrisel (commented):
Quick notes based on our drafting meeting discussion:

joelostblom:
Thanks for working on this! I think these clarification in the docs and the updated behavior would be really helpful to reduce ambiguity. I noticed that in https://github.com/scikit-learn/scikit-learn/issues/26148#issuecomment-1536979006 point 3 there was a mention that the new NumPy Generators would support a different behavior than the old RandomState objects (if I understood correctly). I didn't see a mention of that distinction in this proposal, is this difference still planned to be implemented?

PR #90, opened 2024-03-01
SLEP023: Callback API

jeremiedbb:
Proposal for an API for callbacks in scikit-learn.

The corresponding implementation is being worked in the ``callbacks`` feature branch.
The https://github.com/scikit-learn/scikit-learn/pull/33322 PR keeps an updated diff between ``callbacks`` and ``main`` and is meant to be seen as the implementation of this SLEP in scikit-learn.

### DIFF
--- a/slep023/proposal.rst
+++ b/slep023/proposal.rst
@@ -1,0 +1,144 @@
+.. _slep_023:
+
+=====================
+SLEP023: Callback API
+=====================
+
+:Author: Jérémie du Boisberranger
+:Status: Draft
+:Type: Standards Track
+:Created: 2024-03-01
+
+Abstract
+--------
+
+This SLEP proposes an API to allow users to register callbacks to be called at
+specific points during the training of scikit-learn estimators.
+
+Motivation
+----------
+
+The current scikit-learn API provides a very limited way for users to inspect the steps
+of the training process (`fit`) of an estimator, even less a composition of several
+estimators and meta-estimators. Setting a verbosity level provides some information and
+only a few estimators expose a few public attributes containing some information
+accumulated during training (e.g. `HistGradientBoostingClassifier.train_score_`), and
+that's pretty much it.
+
+Some use cases that have been requested many times on the scikit-learn issue tracker
+which are not possible to achieve with the current API are for instance Progress bars,
+Structured logging, Monitoring, Snapshots, etc. It would also provide a generic and
+consistent API for early stopping, which is currently implemented differently in only a
+few estimators. For such widely requested features, scikit-learn would provide builtin
+callbacks, but users would also be able to implement their own callbacks for less common
+use cases.
+
+By providing a way to gather information during the training process, effectivley making
+scikit-learn implementation of machine learning algorithms more transparent, the
+callback API would also bring a lot of value for educational, testing and debugging
+purposes.
+
+Detailed description
+--------------------
+
+This section discusses the main challenges and the proposed solutions for the callback
+API to be generic and flexible enough despite the wide diversity of scikit-learn
+estimators and meta-estimators.
+
+Callbacks and hooks
+~~~~~~~~~~~~~~~~~~~
+
+This SLEP proposes to define a callback as an object containing methods called at
+specific points of `fit` (hooks). To keep the scope of this SLEP reasonable, only hooks
+called during `fit` are discussed here, but the API could be extended to other methods
+in the future.
+
+One hook is called at the beginning of `fit` (`on_fit_begin`) and one at the
+end (`on_fit_end`) to allow callback initialization and cleanup. A third hook is called
+at the end of each iteration (`on_fit_iter_end`), where an iteration is roughly defined
+as any loop step from the outermost loop in the `fit` method down to a loop over the
+full dataset or batch for online algorithms (e.g. `MiniBatchKMeans`).
+
+Due to the diversity of the estimators in scikit-learn, it's not realistic to define
+specific hooks for each of these kind of iteration, which would end up with estimator
+specific hooks, defeating the purpose of a generic API.
+
+Computation tree
+~~~~~~~~~~~~~~~~
+
+The computation steps of an estimator can be seen as a tree, where the root is the
+beginning of `fit` and nested loops are the different depth levels, each loop step
+being a node. The leaves are the steps of the innermost loop over the full dataset.
+Based on the above section, the `on_fit_iter_end` hook is thus called at each node
+of this tree.
+
+Callbacks need to act differently depending on the depth they are called at. This SLEP
+thus proposes that every estimator hold its computation tree as an attribute, such that
+callbacks can know at which node they are called and act accordingly. Because the tree
+depends on the value of the estimator's hyperparameters, this tree structure is created
+dinamically at the beginning of `fit`. When estimators are nested, the trees are merged,
+the root of the inner estimator being a leaf of the outer estimator.
+
+Performance
+~~~~~~~~~~~
+
+It's inevitable that callbacks will have a performance cost, especially when called
+within cython nogil code. The most important thing is to make sure is that when no
+callbacks are registered the performance is not affected (not acquiring the GIL for
+instance).
+
+In addition, some information given to the `on_fit_iter_end` hook might be expensive to
+compute but is not needed by all callbacks. This SLEP proposes that callbacks must
+explicitly request the information which is expensive to compute, and that the estimator
+gives this information to the hook in a lazy way.
+
+Callback propagation
+~~~~~~~~~~~~~~~~~~~~
+
+When composing estimators and meta-estimators, some callbacks are intuitively expected
+to be applied to all of them. For instance progress bars should be able to report the
+progress of all parts of such compositions. This SLEP proposes that some callbacks have
+the property of being automatically propagated to inner estimators, such that they only
+need to be registered on the outermost estimator.
+
+Parallelism
+~~~~~~~~~~~
+
+Many scikit-learn estimators use multiprocessing or multithreading which can make the
+design of callbacks more complex, because callbacks don't share their state between
+processes. The file system or `multiprocessing.Manager` objects  for instance should be
+used to overcome this issue.
+
+Implementation
+--------------
+
+An implementation of this SLEP is being developed in :pr:`27663`, which targets the
+`callbacks` feature branch. In this PR, only the base infrastructure and one builtin
+callback (progress bars) are implemented. Other callbacks and documention will be
+added incrementally in follow-up PRs.
+
+Discussion
+----------
+
+The goal of this SLEP is to provide a common solution for a wide range of use cases
+discussed in many long-standing issues and PRs, going back to :issue:`78`. Here are
+some related issues and PRs: :issue:`4863`, :issue:`7574`, :issue:`8433`,
+:issue:`8994`, :issue:`9136`, :issue:`10489`, :issue:`10973`, :issue:`12325`,
+:issue:`14338`, :issue:`14531`, :issue:`16118`, :issue:`18507`, :issue:`18748`,
+:issue:`18773`, :issue:`20127`, :issue:`20668`, :issue:`23156`, :issue:`24524`,
+:issue:`25187`, :issue:`26395`, :issue:`26494`, :issue:`26532`, :pr:`1171`, :pr:`3817`,
+:pr:` 8317`, :pr:`16925`, :pr:`22000`.
+
+References and Footnotes
+------------------------
+
+.. [1] Each SLEP must either be explicitly labeled as placed in the public
+   domain (see this SLEP as an example) or licensed under the `Open Publication
+   License`_.
+
+.. _Open Publication License: https://www.opencontent.org/openpub/
+
+Copyright
+---------
+
+This document has been placed in the public domain. [1]_
--- a/slep023/proposal.rst
+++ b/slep023/proposal.rst
@@ -127,7 +127,7 @@ some related issues and PRs: :issue:`4863`, :issue:`7574`, :issue:`8433`,
 :issue:`14338`, :issue:`14531`, :issue:`16118`, :issue:`18507`, :issue:`18748`,
 :issue:`18773`, :issue:`20127`, :issue:`20668`, :issue:`23156`, :issue:`24524`,
 :issue:`25187`, :issue:`26395`, :issue:`26494`, :issue:`26532`, :pr:`1171`, :pr:`3817`,
-:pr:` 8317`, :pr:`16925`, :pr:`22000`.
+:pr:`8317`, :pr:`16925`, :pr:`22000`.
 
 References and Footnotes
 ------------------------
--- a/slep023/proposal.rst
+++ b/slep023/proposal.rst
@@ -101,13 +101,13 @@ To add callback support to an estimator, scikit-learn provides three components:
 
   The `CallbackContext` class exposes the following methods:
 
-  - `eval_on_fit_task_begin`, to evaluate the callbacks at the beginning of the task.
-  - `eval_on_fit_task_end`, to evaluate the callbacks at the end of the task.
+  - `call_on_fit_task_begin`, to call the callbacks at the beginning of the task.
+  - `call_on_fit_task_end`, to call the callbacks at the end of the task.
   - `subcontext`, to create a child `CallbackContext` for a subtask.
   - `propagate_callbacks`, to propagate the callback context and the auto-propagated
     callbacks from a meta-estimator to its sub-estimators.
 
-  When a callback is evaluated for a given task, it receives the corresponding context
+  When a callback is called for a given task, it receives the corresponding context
   which exposes attributes that provide information about the task being executed to be
   used by the callback to adapt its behavior. It also receives all the information that
   the estimator is able to provide about the state of the fitting process at this task.
@@ -137,22 +137,18 @@ A typical implementation of callback support in an estimator would look like thi
             callback_ctx = self._init_callback_context(
                 task_name="fit", max_subtasks=self.n_iter
             )
-            callback_ctx.eval_on_fit_task_begin()
+            callback_ctx.call_on_fit_task_begin()
 
             for i in range(self.n_iter):
                 subcontext = callback_ctx.subcontext(
                     task_name=f"iteration {i}", task_id=i
-                ).eval_on_fit_task_begin(
-                  data={"X_train": X, "y_train": y},
-                )
+                ).call_on_fit_task_begin(X=X, y=y)
 
                 # <computation part of the estimator>
 
-                subcontext.eval_on_fit_task_end(
-                    data={"X_train": X, "y_train": y},
-                )
+                subcontext.call_on_fit_task_end(X=X, y=y)
 
-            callback_ctx.eval_on_fit_task_end()
+            callback_ctx.call_on_fit_task_end()
 
             return self
 
@@ -165,8 +161,8 @@ Callbacks must implement the following protocol:
 
     class FitCallback(Protocol):
         def setup(self, context): ...
-        def on_fit_task_begin(self, context, *, data=None, metadata=None, fitted_estimator=None): ...
-        def on_fit_task_end(self, context, *, data=None, metadata=None, fitted_estimator=None): ...
+        def on_fit_task_begin(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
+        def on_fit_task_end(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
         def teardown(self, context): ...
 
 The `setup` and `teardown` hooks are called at the beginning and end of `fit` and should
@@ -178,17 +174,18 @@ The keyword arguments received by the `on_fit_task_begin` and `on_fit_task_end`
 hooks contain all the available information provided by the estimator about the state of
 the fitting process at this task:
 
-- "data": a dictionary containing the training and validation data.
+- "X": the training data.
+
+- "y": the training target.
 
 - "metadata": a dictionary containing training and validation metadata, e.g. sample
-  weights.
+  weights, `X_val`, `y_val`, etc.
 
 - "fitted_estimator": an estimator instance that is ready to predict, transform, etc.
   as if `fit` stopped at the end of this task.
 
-Note that some estimators may not be able to provide all of these information for every
-task. That list is likely to be extended in the future so callbacks hooks should be
-tolerant to missing keys.
+Note that some estimators may not be able to provide values for all of these keys for
+every task.
 
 Auto-propagated callbacks must implement a small extension of this protocol:
 
@@ -211,11 +208,11 @@ estimators and meta-estimators.
 Protocol extensions
 ~~~~~~~~~~~~~~~~~~~
 
-To keep the scope of this SLEP reasonable, it only considers callbacks evaluated during
+To keep the scope of this SLEP reasonable, it only considers callbacks called during
 `fit` but the API could be extended to other methods (e.g. `predict` or `transform`) or
 unbound functions (e.g. `cross_validate`) in the future.
 PR `#33404<https://github.com/scikit-learn/scikit-learn/pull/33404>`_ for instance
-proposes a new protocol for callback evaluation in unbound functions.
+proposes a new protocol for callback support in unbound functions.
 
 Task granularity
 ~~~~~~~~~~~~~~~~
--- a/slep023/proposal.rst
+++ b/slep023/proposal.rst
@@ -21,17 +21,17 @@ Motivation
 The current scikit-learn API provides a very limited way for users to inspect the steps
 of the training process (`fit`) of an estimator, even less a composition of several
 estimators and meta-estimators. Setting a verbosity level provides some information and
-only a few estimators expose a few public attributes containing some information
-accumulated during training (e.g. `HistGradientBoostingClassifier.train_score_`), and
-that's pretty much it.
+only a few estimators expose public attributes containing some information accumulated
+during training (e.g. `HistGradientBoostingClassifier.train_score_`), and that's pretty
+much it.
 
 Some use cases that have been requested many times on the scikit-learn issue tracker
-which are not possible to achieve with the current API are for instance Progress bars,
-Structured logging, Monitoring, Snapshots, etc. It could also provide a generic and
-consistent API for early stopping, which is currently implemented differently in only a
-few estimators. For such widely requested features, scikit-learn would provide builtin
-callbacks, but users would also be able to implement their own callbacks for less common
-use cases.
+which are not possible to achieve with the current API are for instance progress bars,
+structured logging, metric monitoring, snapshots, etc. The callback API could also
+provide a generic and consistent API for early stopping, which is currently implemented
+differently in only a few estimators. For such widely requested features, scikit-learn
+will provide builtin callbacks, but users would also be able to implement their own
+callbacks for less common use cases.
 
 By providing a way to gather information during the training process, effectivley making
 scikit-learn implementation of machine learning algorithms more transparent, the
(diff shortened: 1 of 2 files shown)


slep023/proposal.rst
```diff
@@ -0,0 +1,144 @@
...
+----------
+
+The goal of this SLEP is to provide a common solution for a wide range of use cases
+discussed in many long-standing issues and PRs, going back to :issue:`78`. Here are
+some related issues and PRs: :issue:`4863`, :issue:`7574`, :issue:`8433`,
+:issue:`8994`, :issue:`9136`, :issue:`10489`, :issue:`10973`, :issue:`12325`,
+:issue:`14338`, :issue:`14531`, :issue:`16118`, :issue:`18507`, :issue:`18748`,
+:issue:`18773`, :issue:`20127`, :issue:`20668`, :issue:`23156`, :issue:`24524`,
+:issue:`25187`, :issue:`26395`, :issue:`26494`, :issue:`26532`, :pr:`1171`, :pr:`3817`,
+:pr:` 8317`, :pr:`16925`, :pr:`22000`.
```
glemaitre, line 130:
```suggestion
:pr:`8317`, :pr:`16925`, :pr:`22000`.
```

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+To add callback support to an estimator, scikit-learn provides three components:
+
+- a `CallbackSupportMixin` class that the estimator should inherit from. This mixin
+  exposes the `set_callbacks` method.
+
+- a `with_callback_context` decorator to wrap the `fit` method of the estimator. This
+  decorator takes care of the setup and teardown of callbacks and makes sure they are
+  properly executed even if an exception is raised during `fit`.
+
+- a `CallbackContext` class that is used to manage the callbacks during `fit`. An
```
thomasjpfan: From reading this doc, it hard to tell what a callback like `Progress` do with the context in `on_fit_task_end`. What is the public interface for `CallbackContext ` that `Progress` should use to show the progress bar?

thomasjpfan (commented):
Thanks for the update! I included some questions about the API.

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+                <computation part of the estimator>
+
+                subcontext.eval_on_fit_task_end(
+                    estimator=self,
+                    context=subcontext,
+                )
+
+            return self
+
+Design considerations
```
StefanieSenger: I think it would be beneficial to add here, what happens if callbacks need additional dependencies (as `ProgressBar` needs `rich`) and if (that) callback dependencies should always be optional.

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+        def on_fit_begin(self, estimator)
+        def on_fit_task_end(self, estimator, context, **kargs)
+        def on_fit_end(self, estimator, context)
+
+The `on_fit_begin` hook is called at the beginning of `fit` and can be used to
+initialize the callback. The `on_fit_end` hook is called at the end of `fit` and can be
+used to cleanup the callback. The `on_fit_task_end` hook is called at the end of each
+task performed during `fit`. Usually tasks correspond to iterations of loops during
+`fit` and nested loops correspond to nested tasks, but in general a task can be any
+unit of work defined by the estimator.
```
StefanieSenger: If the sub tasks don't have a start and only an end, does that mean that the start of the first sub task is right after the start of `fit` and the beginning of the second sub task is right after the first sub task has ended? If it is like this, I imagine there is no much leeway for `on_fit_task_end` and

> "in general a task can be any unit of work defined by the estimator."

can be defined more strictly?

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Callback hooks
+~~~~~~~~~~~~~~
+
+To keep the scope of this SLEP reasonable, it only considers callbacks acting during
+`fit` but the API could be extended to other methods in the future.
+
+In addition, the smallest task granularity considered in this SLEP are tasks dealing
+with the full dataset. For instance the `on_fit_task_end` hook is called at the end of
+a loop iterating over the full dataset but not at the end of each step of such loop.
```
StefanieSenger: Does this mean that `ProgressBar` would not show progress on `cross_validate` and `GridSearchCV` for example?

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Performance
+~~~~~~~~~~~
+
+It's inevitable that callbacks will have a performance cost, especially when called
+within cython nogil code. The most important thing is to make sure is that when no
```
StefanieSenger:
```suggestion
within cython nogil code. The most important thing is to make sure that when no
```

slep023/proposal.rst
jeremiedbb: The context has a few public attributes. They may not be well documented in https://github.com/scikit-learn/scikit-learn/pull/28760 yet. I'll update the PR and this SLEP to better document that part

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+the property of being automatically propagated to inner estimators, such that they only
+need to be registered on the outermost estimator.
+
+The callback protocol
+~~~~~~~~~~~~~~~~~~~~~
+
+Callbacks must implement the following protocol::
+
+    class Callback(Protocol):
+        def on_fit_begin(self, estimator)
```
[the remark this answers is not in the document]
jeremiedbb: So currently it's not needed and I think I never needed it for the other callbacks I prototyped. So I'd say that we should not add args that we know wont be used.
To be precise, everything interesting happens at the end of tasks (on_fit_task_end and on_fit_end), and that where the context is useful. The on_fit_begin is really just for setting stuff up.

Also, the slep matches the current implementation. If we find that we need it here, we'll update the implementation and the slep accordingly.

jeremiedbb:
Thanks for the feedback everyone. One thing I want to precise is that the state of the slep is not final yet. It currently matches the implementation of https://github.com/scikit-learn/scikit-learn/pull/28760 to provide a general vision of the design. However https://github.com/scikit-learn/scikit-learn/pull/28760 targets the `callbacks` feature branch (not `main`) and the goal is to update the slep as we iterate in the `callback` branch and only merge the slep and call for a vote when we're ready to merge main with the callback branch.

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+        def on_fit_task_end(self, estimator, context, **kargs)
```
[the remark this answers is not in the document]
jeremiedbb: another motivation for me for using `**kwargs` is that all callbacks will use a different subset of keys. For instance the progressbar callback doesn't use any of the keys so it would be weird to have
```
class ProgressBar:
    def on_fit_task_end(self, estimator, context, data, stopping_criterion, tol, ...)
```
with all these args after context ignored.

slep023/proposal.rst
```diff
@@ -0,0 +1,235 @@
...
+
+- "from_reconstruction_attributes": an estimator instance that is ready to predict,
+  transform, etc ... estimator as if the fit stopped at the end of this task.
+
+- "fit_state": a dictionary containing model specific quantities updated during fit.
+  This is not meant to be used by generic callbacks but by a callback designed for a
+  specific estimator instead.
+
+Note that all these keys are optional and that estimators are not required to provide
+all of them and within a single estimator different tasks may provide different keys.
```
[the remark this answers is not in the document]
StefanieSenger: Oh thanks! I had totally misunderstood the sentence. 🤦

slep023/proposal.rst
```diff
@@ -0,0 +1,235 @@
...
+The current scikit-learn API provides a very limited way for users to inspect the steps
+of the training process (`fit`) of an estimator, even less a composition of several
+estimators and meta-estimators. Setting a verbosity level provides some information and
+only a few estimators expose a few public attributes containing some information
+accumulated during training (e.g. `HistGradientBoostingClassifier.train_score_`), and
+that's pretty much it.
+
+Some use cases that have been requested many times on the scikit-learn issue tracker
+which are not possible to achieve with the current API are for instance Progress bars,
+Structured logging, Monitoring, Snapshots, etc. It could also provide a generic and
```
lorentzenchr: Is there a reason for the capital letters here "Progress bar", "Structured logging", ..?

slep023/proposal.rst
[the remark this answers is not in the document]
jeremiedbb: In the latest iteration, the protocol has been slightly modified and we pass the context to all hooks. On the other hand we don't pass the estimator to any hook since it's now accessible as an attribute of the context.

I'm going to resolve this conversation since it's about outdated api. Don't hesitate to open a new one if you find issues with the latest version of the protocol.

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+~~~~~~~~~~~~~~~~~~~~~
+
+Callbacks must implement the following protocol::
+
+    class Callback(Protocol):
+        def on_fit_begin(self, estimator)
+        def on_fit_task_end(self, estimator, context, **kargs)
+        def on_fit_end(self, estimator, context)
+
+The `on_fit_begin` hook is called at the beginning of `fit` and can be used to
```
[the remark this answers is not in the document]
jeremiedbb: > Should callbacks follow the convention where the attributes defined in on_fit_begin be suffixed with , i.e. self.my_state?

After some more thoughts, given that such attributes are expected to be shared accrossed processes, e.g. using managers, the attributes only contains proxies to the actual data structure that lives in the manager process. Moreover, they can be difficult to parse since they aggregate data from different levels of tasks in possibly different estimators and meta-estimators. So instead we think that it's more user friendly that the callback provides methods to retreive a well formatted version of the collected data after the run.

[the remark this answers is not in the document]
jeremiedbb: > Should the callback be "clonable", through sklearn.base.clone?

In the latest version, ``clone`` passes the callbacks to the clone by reference.
- In meta-estimators, we want that a single instance of a callback can be registered on all clones of a same estimator.
- In single- or multi-threaded setting it's better to not copy the callback from a memory usage point of view since it can accumulate possibly large amount of data.
- In multi-processing setting, each worker process effectively receives a copy of the callback. Hence the need for callbacks to use process-aware objects, e.g. from managers.

Anyway I think that the precise behavior of ``clone`` w.r.t this SLEP can be seen as an implementation detail.

slep023/proposal.rst
jeremiedbb: So in the end we decided to add a on_fit_task_begin hook. It will be needed when we add callbacks that rely on the difference between the beginning and the end of a task, e.g. memory profiling.

It could also allow a simpler implementation of the progressbars, since we can now start the progressbars at the start of the task. This is low priority since it already work but we can keep that in mind for later.

I'm going to resolve this conversation then since it's about outdated api.

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+A typical implementation of callback support in an estimator would look like this::
+
+    from sklearn.base import BaseEstimator
+    from sklearn.callback import CallbackSupportMixin, with_callback_context
+
+    class MyEstimator(CallbackSupportMixin, BaseEstimator):
+
+        @with_callback_context
+        def fit(self, X, y):
+            callback_ctx = self._callback_fit_ctx
```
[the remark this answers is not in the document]
jeremiedbb: We kind of went for this in the last version. You initialize the callback context using ``self._init_callback_context()``.

It's orthogonal to the decorator though. The purpose of the decorator is to provide a convenient helper to properly tear down the callbacks after fit in a ``try finally`` block to ensure that it's done even if fit errors.

Using the decorator is not mandatory. Instead you can indent your whole fit body in a try finally block and manually tear down the callbacks in the finally. We find that adding 1 indentation level to all fit methods is a good reason to provide that decorato :)

slep023/proposal.rst
jeremiedbb: no :) fixed

slep023/proposal.rst
```diff
@@ -0,0 +1,235 @@
...
+
+Callbacks are objects exposed in the `sklearn.callbacks` module. They can be registered
+to an estimator using their `set_callbacks` method::
+
+    from sklearn.linear_model import LogisticRegression
+    from sklearn.callbacks import ProgressBar
+
+    clf = LogisticRegression()
+    callback = ProgressBar()
+    clf.set_callbacks(callback)
```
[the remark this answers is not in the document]
jeremiedbb: That can indeed be a usage improvement. It doesn't need to part of the SLEP though I believe.
Also, note that callbacks have parameters, even ProgressBar, so a string would only provide a callback with default or pre-chosen parameters.

slep023/proposal.rst
```diff
@@ -0,0 +1,235 @@
...
+
+This sections describes the proposed public interfaces for the callback API. They are
+divided into three subsets which target different kind of users: the end users of
+scikit-learn, the third-party developers implementing custom estimators and the
+third-party developers implementing custom callbacks.
+
+Using callbacks
+~~~~~~~~~~~~~~~
+
+Callbacks are objects exposed in the `sklearn.callbacks` module. They can be registered
```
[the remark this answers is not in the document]
jeremiedbb: A can think of a couple reasons
- We want to allow third-party callbacks. So a string can't be passed, unless scikit-learn provides a callback regisrty mechanism, but we lose the "simple" API goal.
- Callbacks are object that need to stay alive after ``fit`` so that the user can access its accumulated data. For instance a callback logging metric values during fit.

> Especially for GLMs, it would have been quite natural, IMO, to pass the loss as an object instantiated with its own arguments, e.g. TweedieLoss(power=1).

Believe me I'm on your side here 😄.

jeremiedbb:
Hi everyone,

I've updated the SLEP with the latest version of the implementation that is happening in the ``callbacks`` feature branch.
We believe that this version is (almost) ready for a call for a vote. Our goal is to merge this PR and call for the vote by the end of the week (3-20-2026).

- PR https://github.com/scikit-learn/scikit-learn/pull/33322 keeps an up-to-date diff between ``callbacks`` and ``main`` to showcase how the implementation of this SLEP will look like in scikit-learn. It currently includes the callback framework and a callback for progressbars.

- A few ongoing PRs (they may not all be up-to-date with the latest version of the SLEP as the time of writing) on the ``callbacks`` branch can also help reviewing this SLEP:

    - https://github.com/scikit-learn/scikit-learn/pull/33407 Implements a callback to monitor a given metric during fit. We aim to merge this PR before merging the SLEP because, belonging to the second kind of callbacks, it might still bring small modifications to the SLEP.
    - https://github.com/scikit-learn/scikit-learn/pull/33423 Adds documentation for scikit-learn and third-party developers to add callback support to an estimator. A separate PR that adds documentation for end users will be open soon.
    - https://github.com/scikit-learn/scikit-learn/pull/33533 implements callback support in *SearchCV objects to illustrate of endorsing this SLEP would look like concretely in scikit-learn estimators. We are going to open a few other PRs like that for e.g. Pipeline, LogisticRegression, ... to propose a better overview.

- https://github.com/scikit-learn/scikit-learn/pull/33404 proposes an extension of the protocol to be able to add callback support in unbound functions like ``cross_validate``. It's not meant to be part of this SLEP but to show how we can extend the scope of the callback framework without modifying the existing API.

The reason for this very tight schedule is that we're hoping to have a minimal support for callbacks (in a few key (meta)estimators) in the next scikit-learn release, due mid May. Given that the vote will last 1 month, and that we need some time after the vote if it's accepted to finalize and merge the different PRs in the ``main`` branch, there's not much room left.

We hope that the latest version of the SLEP answers the questions that were raised so far. If you find some time in the coming days to take a new look at the SLEP, we'll very much appreciate all feedback that will help to improve it and increase its chance of being accepted.

slep023/proposal.rst
```diff
@@ -0,0 +1,313 @@
...
+A typical implementation of callback support in an estimator would look like this:
+
+.. code-block:: python
+
+    from sklearn.base import BaseEstimator
+    from sklearn.callback import CallbackSupportMixin, with_fit_callbacks
+
+    class MyEstimator(CallbackSupportMixin, BaseEstimator):
+
+        @with_fit_callbacks
```
thomasjpfan: Will there be other non-fit decorators, like `with_transfrom_callbacks`? If so what do you think of `with_callbacks("fit")`?

Do you think it'll end up as follows and we'll have a `with_transform_callbacks`?

```python
class TransformCallback(Protocol):
    def on_transform_task_begin(...)
    def on_transform_task_end(...)
```

slep023/proposal.rst
```diff
@@ -0,0 +1,313 @@
...
+        def fit(self, X, y):
+            callback_ctx = self._init_callback_context(
+                task_name="fit", max_subtasks=self.n_iter
```
thomasjpfan, line 139: Does the task name have to be "fit" or can it be anything?

slep023/proposal.rst
```diff
@@ -0,0 +1,313 @@
...
+            )
+            callback_ctx.eval_on_fit_task_begin()
```
thomasjpfan: Using the `eval` prefix in a ML library feels like it's doing some ML evaluation, but the callback could just be doing some setup. Can it be `run_on_fit_task_begin` or `fit_task_begin`?

slep023/proposal.rst
```diff
@@ -0,0 +1,313 @@
...
+The callback protocol
+~~~~~~~~~~~~~~~~~~~~~
+
+Callbacks must implement the following protocol:
+
+.. code-block:: python
+
+    class FitCallback(Protocol):
+        def setup(self, context): ...
+        def on_fit_task_begin(self, context, **kwargs): ...
```
thomasjpfan: I like `on` more than `eval`. (This is inconsistent with the section above)

slep023/proposal.rst
```diff
@@ -0,0 +1,313 @@
...
+design of callbacks more complex, because callbacks don't share their state between
+processes. The file system or `multiprocessing.Manager` objects for instance should be
+used to overcome this issue.
+
+Non-callback-aware meta-estimators
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Callbacks can be registered on estimators that are passed to meta-estimators or unbound
+functions that do not support callbacks. In this case, callbacks should not break the
+workflow but might not provide full value and perform their work sub-optimally.
```
thomasjpfan: Does that mean the callback needs to be written very defensively? For example, the callback will check for all the `kwargs` keys in `on_fit_task_begin/on_fit_task_end` and make sure it does not error.

slep023/proposal.rst
```diff
@@ -0,0 +1,313 @@
...
+being automatically propagated down to sub-estimators, such that they only need to be 
+registered on the outermost meta-estimator:
+
+.. code-block:: python
+
+    from sklearn.linear_model import LogisticRegression
+    from sklearn.model_selection import GridSearchCV
+    from sklearn.callbacks import EarlyStopping, ProgressBar
+
+    clf = LogisticRegression().set_callbacks(EarlyStopping())
```
thomasjpfan, line 84: Is `ProgressBar` passed into `LogisticRegression`? If so, what happens if the inner estimator supports the callbacks, but it does not pass all the `kwargs` needed for `ProgressBar` to work?

slep023/proposal.rst
```diff
@@ -0,0 +1,313 @@
...
+  defined by the estimator. There is one `CallbackContext` for each task. Tasks (and
+  therefore callback contexts) have a natural tree structure, where each task can be
+  decomposed into subtasks and so on, with the root task being the whole `fit` task.
+  Usually tasks correspond to iterations of loops during `fit` and nested loops
+  correspond to nested tasks, but in general a task can be any unit of work defined by
+  the estimator.
+
+  The `CallbackContext` class exposes the following methods:
+
+  - `eval_on_fit_task_begin`, to evaluate the callbacks at the beginning of the task.
```
thomasjpfan: How will we handle `partial_fit`?

thomasjpfan (commented):
Thank you for the update! I left some of my thoughts & questions.

slep023/proposal.rst
jeremiedbb: > If so what do you think of with_callbacks("fit")?

Good idea, it's better extendable to future extensions.
It could even be ``with_callbacks()`` and dispatch according to the name of the method it received.

slep023/proposal.rst
jeremiedbb: It can be anything, see the subtask is named "iteration" below. It's meant to be a informative short description of the task that the callback can retrieve and display (e.g. progress bars) or store (e.g. metric monitoring).

slep023/proposal.rst
jeremiedbb: I can see the confusion with model evaluation.
``run_on_fit_task_begin`` would be okay for me 👍 It's explicit that it runs the ``on_fit_task_begin`` hook of the callbacks (unlike just `fit_task_begin`)

slep023/proposal.rst
jeremiedbb: - ``on_fit_task_begin`` is the name of the method of the callback. This is described in the Callback Protocol section that targets developers of Callback objects. The ``on`` reflects that this method is called when a specific even happens, here the beginning of a fit task.
- ``eval/run_on_fit_task_begin`` is the name of the method of the CallbackContext class, that calls the former for all callbacks registered on the estimator. This is described in the section that targets developer who want support callback in their estimators.

slep023/proposal.rst
jeremiedbb: > Is ProgressBar passed into LogisticRegression?

In this example yes because ProgressBar is an auto-propagated callback which means it's propagated to nested sub-estimators.

> If so, what happens if the inner estimator supports the callbacks, but it does not pass all the kwargs needed for ProgressBar to work?

The estimator passes all the information it can to the callbacks. It's a subset of keyword args from a defined set of possible args ("data", "metadata", "fitted_estimator" so far but may be extended in the future). It's not expected that all estimators pass all of them at every task level.
The callback can then act differently at different tasks based on the args it received. We also make sure that you can implement callback support in an estimator incrementally. You can start by implementing minimal support, passing no info, and then later make a PR to pass the "data" info for instance.

In case of progress bars, it's particular because it doesn't need any of these args so it will just work on every estimator that implement the most minimal callback support.

slep023/proposal.rst
jeremiedbb: I think that ``partial_fit`` counts as a fit method, so support for fit callbacks will be added to partial_fit in estimators.
To some extent, it's the same for ``fit_transform`` and ``fit_predict``, which will at some point likely also support predict or transform callbacks.

slep023/proposal.rst
jeremiedbb: No the callback will be guaranteed to receive keys from the defined set of possible keys.

Instead it will be the responsibility of the estimator to pass valid keys. It will be checked by the CallbackContext. The last part, i.e. the check in callback context, is an implementation detail so is not part of the SLEP.

thomasjpfan: > Instead it will be the responsibility of the estimator to pass valid keys. It will be checked by the CallbackContext.

How does the CallbackContext or Estimator know what are all the keys a specific callback will understand? (Let's say there is a third party `MySpecialEarlyStoppingCallback`) From my reading, an estimator will pass everything it can and the callback will adapt to those keys.

I see that estimators can pass a subset of the keys, but it's unclear what the subsets are. Does one need to run code with a `PrintKeysCallback` and print out all the keys to get this information?

jeremiedbb: The set of possible keys is defined once at the scikit-learn level and is the same for all estimators and for all callbacks. That list is described in the Callback Protocol section and currently contains ("data", "metadata", "fitted_estimator").

- All callbacks receive the same keyword args (from that list). Then they use only the ones they need.
- For a given task, an estimator provides all args (from that same list) that make sense for this task.

So from within an estimator we'll have something like:
```python
def fit
     root_context = self._init_callback_context()
     ...
     subcontext = root_context.subcontext().
     # For this task only the data and metadata info can be provided by the estimator.
     subcontext.eval_on_fit_task_begin(
         data = {X, y}
         metada = sample_weight
     )
     ...
```

And the counter part from the callback is
```python
def on_fit_task_begin(self, context, **kwargs):
    data = kwargs.get("data", None)
    metadata = kwargs.get("metadata", None)
    fitted_estimator = kwargs.get("fitted_estimator", None)

    if fitted_estimator:
        # do something...
    else:
        # the task being executed is not relevant for this callback
        # do nothing
```

jeremiedbb: Okay I think I'm starting to see where the confusion comes from.

I wrote the signature as ``def  on_fit_task_begin(context, **kwargs)``. What I probably should write is the exact expected signature ``def on_fit_task_begin(context, *, data, metadata, fitted_estimator)``. And all callbacks must have that signature, then they can do whatever they want with the args they receive.

I wrote it with **kwargs because, when concretely implementing callbacks, we explicitely write the args the callback needs and leave the others as **kwargs. I realise that was a mistake, and will probably be a lot clearer if I update the SLEP with the full signature.

jeremiedbb: One reason I used **kwargs though is that it allows future extension of that list of args in the future without breaking existing callback classes. But that's probably an implementation detail.

jeremiedbb: I updated the SLEP, I hope it's clearer now

slep023/proposal.rst
```diff
@@ -162,18 +165,18 @@ Callbacks must implement the following protocol:
 
     class FitCallback(Protocol):
         def setup(self, context): ...
-        def on_fit_task_begin(self, context, **kwargs): ...
-        def on_fit_task_end(self, context, **kwargs): ...
+        def on_fit_task_begin(self, context, *, data=None, metadata=None, fitted_estimator=None): ...
```
thomasjpfan: Does the signature of the callback still need `**kwargs` at the end? This way third party callback developers will not break when we pass a new kwargs?

jeremiedbb: Not necessarily. We're currently thinking that the context should inspect the signature of the callback hook and only pass the args that are in the signature.

(
It allows us to write signatures with only a subset of these args and be able to passe them lazily to avoid unnecessary performance degradation. It doesn't need to be part of this SLEP though and can be seen as future improvement.
)

So I'd consider that an implementation detail, no ?

slep023/proposal.rst
jeremiedbb: Done. The decorator is now ``with_callbacks()``. When we extend the API to transform, the decorator will be responsible to tear down appropriately based on the method it decorates.

Therefore I'm marking this conversation as resolved

slep023/proposal.rst
jeremiedbb: will be changed in https://github.com/scikit-learn/scikit-learn/pull/33561

slep023/proposal.rst
thomasjpfan: I'm mostly thinking about this SLEP from the view point of a third party callback developer and what they did to know to develop one. To me, the interface between the estimator and the callback is public developer API and not an implementation detail.

> context should inspect the signature of the callback hook and only pass the args that are in the signature.

The current `FitCallback` protocol specifies a finite set of kwargs. Lets say we extend `FitCallback.on_fit_task_begin` to have a new `more_metadata` kwargs, do third party callbacks have to update their code to be compatible? We can have callback common tests to help developers update, which is what we already do every time we add a new requirement to estimator checks.

ogrisel: > Lets say we extend FitCallback.on_fit_task_begin to have a new more_metadata kwargs, do third party callbacks have to update their code to be compatible? 

If we do the signature inspection and only pass values for inspected arguments, the callback developers will not have to update anything and this further allows to not precompute expensive argument values when not needed (e.g. a virtual "fitted" estimator instance that can be scored or snapshot at each iteration of fit).

> We can have callback common tests to help developers update, which is what we already do every time we add a new requirement to estimator checks.

+1 for a standard set of callback checks to assist third party developers and dedicated documentation (but this can come after the SLEP is voted).

jeremiedbb: > To me, the interface between the estimator and the callback is public developer API and not an implementation detail.

I agree, what I consider an implementation detail is what scikit-learn does internally to make sure that third-party callbacks don't break when we extend the signature.

slep023/proposal.rst
```diff
@@ -0,0 +1,317 @@
...
+registered on the outermost meta-estimator:
+
+.. code-block:: python
+
+    from sklearn.linear_model import LogisticRegression
+    from sklearn.model_selection import GridSearchCV
+    from sklearn.callbacks import EarlyStopping, ProgressBar
+
+    clf = LogisticRegression().set_callbacks(EarlyStopping())
+    clf = GridSearchCV(clf, param_grid).set_callbacks(ProgressBar())
```
ogrisel: Maybe setting `max_propagation_depth` explicitly would make it more explicit that `ProgressBar` is propagated to `LogisticRegression`.

```suggestion
    clf = GridSearchCV(clf, param_grid).set_callbacks(
        ProgressBar(max_propagation_depth=1)
    )
```

slep023/proposal.rst
```diff
@@ -0,0 +1,317 @@
...
+
+- "metadata": a dictionary containing training and validation metadata, e.g. sample
+  weights.
+
+- "fitted_estimator": an estimator instance that is ready to predict, transform, etc.
+  as if `fit` stopped at the end of this task.
+
+Note that some estimators may not be able to provide all of these information for every
+task. That list is likely to be extended in the future so callbacks hooks should be
+tolerant to missing keys.
```
ogrisel: I don't follow the logic for tolerance to missing keys. Could you give an example of a situation where we write a callback with a hook that optionally accepts values that are sometimes provided by the callback context initialized by the estimator at fit time?

ogrisel: Or did you mean the opposite? "That list is likely to be extended in the future so callbacks hooks should accept to be provided extra keys.`

But this would not be necessary if the callback context inspects the signature of the hooks implemented by the callback to only provide values for explicitly set keys as discussed in https://github.com/scikit-learn/enhancement_proposals/pull/90/changes#r2945238930.

slep023/proposal.rst
jeremiedbb: Done. We voted on a few propositions and `call` was selected :)

slep023/proposal.rst
```diff
@@ -0,0 +1,314 @@
...
+
+Callbacks must implement the following protocol:
+
+.. code-block:: python
+
+    class FitCallback(Protocol):
+        def setup(self, context): ...
+        def on_fit_task_begin(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
+        def on_fit_task_end(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
+        def teardown(self, context): ...
```
jeremiedbb, line 168: Small adjustment based on the latest development. Explicitly pass X and y instead of data.

I initially thought that `X_val` and `y_val` were not managed by the metadata routing so I had put both in data alongside X and y. But looking back at the `transform_input` param of the pipeline ([see](https://scikit-learn.org/dev/modules/generated/sklearn.pipeline.Pipeline.html#:~:text=transform_input,-list%20of%20str%2C%20default%3DNone)) and at this [release highlight](https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_6_0.html#transforming-data-other-than-x-in-a-pipeline), we realized that X_val and y_val are (and must be) routed using metadata routing.

So data would only contain X and y and it felt more natural and consistant with sklearn usage to directly pass them.

slep023/proposal.rst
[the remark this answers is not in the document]
jeremiedbb: The document has changed quite a bit after this discussion.
The signature is now fully explicit in the SLEP to avoid confusion.
I'm resolving this conversation

slep023/proposal.rst
jeremiedbb: In the end we made sure that callbacks still work in that situation. They may not work at full capacity though, e.g. progress bars overlapping. I added a small section under the "considerations" section to explain that.

I'm going to resolve this conversation then

slep023/proposal.rst
```diff
@@ -0,0 +1,314 @@
...
+being automatically propagated down to sub-estimators, such that they only need to be 
+registered on the outermost meta-estimator:
+
+.. code-block:: python
+
+    from sklearn.linear_model import LogisticRegression
+    from sklearn.model_selection import GridSearchCV
+    from sklearn.callbacks import EarlyStopping, ProgressBar
+
+    clf = LogisticRegression().set_callbacks(EarlyStopping())
```
ogrisel, line 84: Maybe `EarlyStopping` is a bit of a complex case because it would typically require the user to specify which validation data to use, which in turns would require configuring metadata requests to propagate such validation data through meta-estimators (especially pipelines). I would rather like to try to decouple the design of callback propagation from other more advanced considerations related to routing metadata through meta-estimators.

Maybe it's best to using another example such as `Snapshot(folder="./snaphots")` that wouldn't require access to routed metadata.

slep023/proposal.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Callback support in estimators
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To add callback support to an estimator, scikit-learn provides three components:
+
+- a `CallbackSupportMixin` class that the estimator should inherit from. This mixin
+  exposes the `set_callbacks` method.
+
+- a `with_callback_context` decorator to wrap the `fit` method of the estimator. This
```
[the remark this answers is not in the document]
jeremiedbb: I forgot to mention that using the decorator is not required. Instead you can add one level of indentation in all your fit methods to wrap the body of fit in a `try finally` block and manually tear down the callbacks in the finally.
The decorator is here as convenience, and strongly suggested to be used to avoid painful mistakes, e.g. leaving threads alive.

In the long run, the preferred solution would be that scikit-learn has a public `fit` and a private `_fit` where the private fit is only about the optimization problem and the public fit is about the boilerplate. Then the `try finally` would be done by the public fit and we could get rid of the decorator.
This discussion is orthogonal to the callbacks and not happening tomorrow, so in the meantime the decorator is a good enough solution.

slep023/proposal.rst
```diff
@@ -0,0 +1,314 @@
...
+
+        @with_callbacks
+        def fit(self, X, y):
+            callback_ctx = self._init_callback_context(
+                task_name="fit", max_subtasks=self.n_iter
+            )
+            callback_ctx.call_on_fit_task_begin()
+
+            for i in range(self.n_iter):
+                subcontext = callback_ctx.subcontext(
```
ogrisel: Maybe we could rename `subcontext` to `callback_subctx` for consistency?

slep023/proposal.rst
```diff
@@ -0,0 +1,314 @@
...
+    from sklearn.callback import CallbackSupportMixin, with_callbacks
+
+    class MyEstimator(CallbackSupportMixin, BaseEstimator):
+
+        @with_callbacks
+        def fit(self, X, y):
+            callback_ctx = self._init_callback_context(
+                task_name="fit", max_subtasks=self.n_iter
+            )
+            callback_ctx.call_on_fit_task_begin()
```
ogrisel: Shouldn't we also pass `X` and `y` here?

slep023/proposal.rst
```diff
@@ -0,0 +1,314 @@
...
+                subcontext.call_on_fit_task_end(X=X, y=y)
+
+            callback_ctx.call_on_fit_task_end()
+
+            return self
+
+The callback protocol
+~~~~~~~~~~~~~~~~~~~~~
+
+Callbacks must implement the following protocol:
```
ogrisel: Please link to the Python documentation for protocols because not everybody is familiar with them and how they work.

slep023/proposal.rst
```diff
@@ -0,0 +1,314 @@
...
+
+.. code-block:: python
+
+    class AutoPropagatedCallback(FitCallback, Protocol):
+        @property
+        def max_propagation_depth(self): ...
+
+`max_propagation_depth` defines the maximum nesting level of sub-estimators to propagate
+the callback to.
+
```
ogrisel, line 203: I think we should give add an example section here with:

- code snippets to instantiate some fake meta-estimators, estimators and callbacks (auto-propagated or not)
- trace of all invocations to `setup`, `on_fit_task_begin/end` hooks, `teardown`.

Otherwise, it's not easy to infer the dynamic effects of a static code description.

ogrisel (commented):
Here is a pass of feedback.

slep023/proposal.rst
jeremiedbb: Yes I already removed that line :)

slep023/proposal.rst
jeremiedbb: We can, we don't have to. The estimator decides what it does :)
I left some of them empty to show that but if you find that more confusing maybe better to fill them all

slep023/proposal.rst
jeremiedbb: That's true but
- we don't talk about metadata routing here, it's just an example to illustrate
- EarlyStopping feels like a more common use case, know by everybody

slep023/proposal.rst
jeremiedbb: done

slep023/proposal.rst
jeremiedbb: I added example traces. Let me know if that's what you had in mind.

slep023/proposal.rst
```diff
@@ -0,0 +1,399 @@
...
+.. code-block:: python
+
+    class FitCallback(Protocol):
+        def setup(self, context): ...
+        def on_fit_task_begin(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
+        def on_fit_task_end(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
+        def teardown(self, context): ...
+
+The `setup` and `teardown` hooks are called at the beginning and end of fit and should
```
FrancoisPgm: Maybe "hooks" should be introduced before being mentioned here.

slep023/proposal.rst
ogrisel: Thanks, this is really useful!

slep023/proposal.rst
ogrisel: I agree, but I find it a bit misleading to suggest that this user code will be enough to get early stopping to work in practice. This simplistic code would only be able to implement early stopping on the training set which nobody wants to do (I believe). To implement early stopping on a validation set, the user would have to pass validation data to the callback one way or another.

slep023/proposal.rst
```diff
@@ -0,0 +1,399 @@
...
+be used to set up and tear down the callback for the estimator. The `on_fit_task_begin`
+and `on_fit_task_end` hooks are called at the beginning and end of each task performed
+during fit, including the root task.
```
ogrisel, line 174: Maybe we could be a bit more explicit about the motivations behind the choice of separating the setup/teardown from the `on_fit_task_begin/end` on the outer fit task.

I think the main motivations are:

- to be able to allocate/free shared resources (e.g. rich datastructures) only once on the outer/root call for auto-propagated callbacks such as the progress bar for meta estimators;
- to separate technical concerns of resource managements from semantic concerns related to reacting to particular estimator's fit method specific event. This latter point will mostly prove useful if/when we later extend the callback protocol to handle other estimator methods (`transform`, `predict`) or other public functions (`cross_validate`, ...). In this case the callback will not have to update their existing `setup`/`teardown` methods but will have the choice to also implement extra logic related to those specific methods.

ogrisel: This is a bit complex to explain, I don't want to make the SLEP overly verbose but in its current state, it seems to lack some motivation behind the way the callback protocol is designed.

ogrisel (commented):
One more comment:

slep023/proposal.rst
```diff
@@ -0,0 +1,399 @@
...
+This section gives example traces of callback hook calls in different scenarios. First,
+consider a callback registered on an estimator with a single level of iterations:
+
+.. code-block:: python
+
+    >>> estimator = MyEstimator(n_iter=10).set_callbacks(MyCallback())
+    >>> estimator.fit(X, y)
+    setup by MaxIterEstimator for fit
+    on_fit_task_begin by MaxIterEstimator for fit
+    on_fit_task_begin by MaxIterEstimator for iteration 0
```
ogrisel: Maybe we could use indentation levels in this log to reflect the depth of the subcontext tree structure.

jeremiedbb: done

slep023/proposal.rst
jeremiedbb: I didn't include it there but I added a dedicated section under the Considerations section where I describe the motivation for having them.

slep023/proposal.rst
jeremiedbb: Done

slep023/proposal.rst
jeremiedbb: So in the end I kept EarlyStopping.
- it's a very common use case. (more than snapshots I believe)
- We have a POC that we can achieve it with this API (including on a validation set)
- Other callbacks that I tried to think about were auto-propagated ones, or ScoreMonitoring that has the same concerns as EarlyStopping. The only one left was snapshot for which I'm not sure yet if it should/could be auto-propagated or not.

I don't think it will matter too much anyway :smile:

slep023/proposal.rst
```diff
@@ -0,0 +1,411 @@
...
+
+.. code-block:: python
+
+    class FitCallback(Protocol):
+        def setup(self, context): ...
+        def on_fit_task_begin(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
+        def on_fit_task_end(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
+        def teardown(self, context): ...
+
+The 4 protocol members are referred to as callback hooks in the rest of this SLEP. The
```
jeremiedbb:
```suggestion
The 4 protocol methods are referred to as callback hooks in the rest of this SLEP. The
```

ogrisel (approved):

jeremiedbb:
Here goes nothing :grimacing:

PR #94, opened 2024-11-06
SLEP25 BaseEstimator.freeze

adrinjalali:
This slep is to address the comments about https://github.com/scikit-learn/scikit-learn/pull/30211 needing a SLEP.

### DIFF
--- a/slep025/proposal.rst
+++ b/slep025/proposal.rst
@@ -45,6 +45,51 @@ The proposal of this SLEP is implemented in [6]_.
 
 This change has no backward compatibility concerns.
 
+
+Implementation
+--------------
+
+It is implemented in [6]_ as::
+
+   class BaseEstimator(...):
+      ...
+      def freeze(self):
+         """Return a :term:`frozen` version of this estimator.
+
+         This method wraps the estimator in a :class:`~sklearn.frozen.FrozenEstimator`
+         object, and returns the frozen object. This method does NOT freeze this object
+         itself.
+
+         The return frozen object will prevent changes to this estimator's instance
+         parameters; however, direct changes to this estimator are still possible.
+         Returns
+         -------
+         frozen : :class:`~sklearn.frozen.Frozen`
+               A frozen version of this estimator.
+         """
+         from sklearn.frozen import FrozenEstimator
+         return FrozenEstimator(self)
+
+
+Alternatives
+------------
+
+1. We could not add the method. The user would always have to manually import
+``FrozenEstimator`` from ``sklearn.frozen``, and wrap their estimator with it.
+
+2. We could provide a method under `skelarn` or `sklearn.base`. Which would mean the
+user would have to import from ``sklearn`` or ``sklearn.base`` to use the method::
+
+    from sklearn.calibration import CalibratedClassifierCV
+    from sklearn.ensemble import RandomForestClassifier
+    from sklearn import freeze
+
+    clf = RandomForestClassifier(n_estimators=25)
+    clf.fit(X_train, y_train)
+    cal_clf = CalibratedClassifierCV(freeze(clf), method="sigmoid")
+    cal_clf.fit(X_valid, y_valid)
+
+
 Copyright
 ---------
 
--- a/slep025/proposal.rst
+++ b/slep025/proposal.rst
@@ -1,0 +1,70 @@
+.. _slep_025:
+
+====================================
+SLEP025: BaseEstimator.freeze method
+====================================
+
+:Author: Adrin Jalali
+:Status: Draft
+:Type: Standard
+:Created: 2024-10-06
+
+Abstract
+--------
+This SLEP proposes adding a ``freeze`` method to the ``BaseEstimator`` class.
+[2]_ added a `sklearn.frozen.FrozenEstimator` class to scikit-learn. [3]_ started
+a discussion on the implications of adding a ``FrozenEstimator`` to scikit-learn.
+
+As a part of that discussion, it became clear that we'd like to use `FrozenEstimator`
+instead of having patterns such as `cv="prefit"` or `prefit=True` in meta-estimators.
+[4]_ and [5]_ are examples of how this change would impact the API.
+
+However, this means user's code would look like the following::
+
+    from sklearn.calibration import CalibratedClassifierCV
+    from sklearn.ensemble import RandomForestClassifier
+    from sklearn.frozen import FrozenEstimator
+
+    clf = RandomForestClassifier(n_estimators=25)
+    clf.fit(X_train, y_train)
+    cal_clf = CalibratedClassifierCV(FrozenEstimator(clf), method="sigmoid")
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
Adding a method to our public API is not something minor: it imposes to all downstream packages that claim to be scikit-learn compatible to implement it.

I do not think that the decision on this should be rushed (without weighing one way or the other)

adrinjalali:
> Adding a method to our public API is not something minor: it imposes to all downstream packages that claim to be scikit-learn compatible to implement it.

@GaelVaroquaux the downstream libraries don't have to do anything as long as they inherit from `BaseEstimator`. And at this point, if somebody doesn't want to inherit from `BaseEstimator`, good luck with them; it's almost impossible to be completely compatible if they don't want to inherit from the right mixins, and we don't really consider not inheriting from the right mixins a supported way for building sklearn compatible estimator.

amueller:
I think it would be nice to add a discussion on the alternatives. Personally I'm not convinced of the benefit of a method over a meta-estimator. This seems like a very rare use-case and adding a method for it seems overkill. I've wanted to bag something more often than freeze it, yet I don't think we wouldn't consider adding a ``.bag()`` method. All meta-estimators could be implemented using a fluid API and then there would be much less imports, but that's not really how sklearn code looks right now.

adrinjalali:
> I think it would be nice to add a discussion on the alternatives. Personally I'm not convinced of the benefit of a method over a meta-estimator.

@amueller this is via a meta-estimator. We have `FrozenEstimator` now included in scikit-learn. This is the implementation of the proposed method:

```py
class BaseEstimator(...):
    ...
    def freeze(self):
        """Return a :term:`frozen` version of this estimator.

        This method wraps the estimator in a :class:`~sklearn.frozen.FrozenEstimator`
        object, and returns the frozen object. This method does NOT freeze this object
        itself.

        The return frozen object will prevent changes to this estimator's instance
        parameters; however, direct changes to this estimator are still possible.
        Returns
        -------
        frozen : :class:`~sklearn.frozen.Frozen`
            A frozen version of this estimator.
        """
        from sklearn.frozen import FrozenEstimator
        return FrozenEstimator(self)
```

amueller:
Thanks for the addition & context. Not very convinced tbh.

adrinjalali:
@adam2392 @ogrisel @thomasjpfan @lorentzenchr @glemaitre @Charlie-XIAO @OmarManzoor  WDYT here? We need to make a decision here to be able to mvoe forward.

slep025/proposal.rst
```diff
@@ -0,0 +1,115 @@
...
+         return FrozenEstimator(self)
+
+
+Alternatives
+------------
+
+1. We could not add the method. The user would always have to manually import
+``FrozenEstimator`` from ``sklearn.frozen``, and wrap their estimator with it.
+
+2. We could provide a method under `skelarn` or `sklearn.base`. Which would mean the
```
glemaitre, line 80:
```suggestion
2. We could provide a method under `sklearn` or `sklearn.base`. Which would mean the
```

glemaitre:
As stated in #29893, I would like to have `freeze` and `clone` to be both either a function or both a method.

I'm pretty much undecided on the matter:

- not having to import could bring a little plus.
- but if the use-case is rare enough (as argued by Andy) and that `clone` or `freeze` are exposed at the top level such that one uses `sklearn.clone(...)` and `sklearn.freeze(...)` then I would be happy with functions.

So I'm probably leaning really slightly towards a function because `clone` is already a function and I don't want to pay the cost of the deprecation.

solegalli:
From my experience as a maintainer of a sklearn compatible package, changing the `BaseEstimator`, does create some work on our side, because if anything else, we need to add some tests and ensure that the new methods work as expected in our transformers as well.

glemaitre:
> we need to add some tests and ensure that the new methods work as expected in our transformers as well

In a perfect world, this should be covered on our side because we should provide the compatibility test.

> does create some work on our side

I acknowledge that it does and the main reason is that we don't have a proper developer API and did not provide public tooling sometimes. I'm hopeful that it will get better in this regard.

adrinjalali:
> From my experience as a maintainer of a sklearn compatible package, changing the BaseEstimator, does create some work on our side, because if anything else, we need to add some tests and ensure that the new methods work as expected in our transformers as well.

You won't need to do anything in almost all cases, definitely in this case, as long as you inherit from `BaseEstimator` @solegalli

lorentzenchr:
> We need to make a decision here to be able to mvoe forward.

There is no need to rush. We can release 1.6 without it. We have `FrozenEstimator`, so we are only talking about convenience.

If we add it, I am in favor of having both, clone and freeze, as methods instead of as functions. That feels more natural for me and would also eliminate my confusion about clone vs copy functions as well as make it easier to customize by 3rd party libs.

About the inclusion of `freeze`, I am currently +0. Certainly not against it, but not yet able to anticipate its need (how often does someone need to freeze something?) and therefore woring about implementing something nobody asked for.

betatim:
What is the "copy functions" thing people are mentioning? Is this `copy.copy` from the Python standard library?

I think I have a preference for `clone` and `freeze` being functions. It doesn't seem like freezing will be a super frequent task for the average user and asymmetry between the two seems weird.

slep025/proposal.rst
```diff
@@ -0,0 +1,115 @@
...
+
+
+Alternatives
+------------
+
+1. We could not add the method. The user would always have to manually import
+``FrozenEstimator`` from ``sklearn.frozen``, and wrap their estimator with it.
+
+2. We could provide a method under `sklearn` or `sklearn.base`. Which would mean the
+user would have to import from ``sklearn`` or ``sklearn.base`` to use the method::
```
betatim, line 81:
```suggestion
2. We could provide a function under `sklearn` or `sklearn.base`. Which would mean the
user would have to import from ``sklearn`` or ``sklearn.base`` to use the fucntion::
```

adrinjalali:
I personally don't see the point of `sklearn.freeze(est)` as opposed to `est.freeze()` where one can do `sklearn.frozen.FrozenEstimator(est)` since it doesn't add enough of a convenience.

Since there isn't enough support, I retract the proposal.

GaelVaroquaux:
My apologies for reacting slowly. I wanted to think this over, and also I'm always multitasking, which does not help.

So, having though a bit about this, I could see an interesting case for a freeze method: if the freeze method would return a different object (which could be based on a copy, or not, of the original object), where the only contract would be that the returned object would have a numerically equivalent predict (or transform), but not fit.

Why would I be excited about such proposal? It would upon the door for lightweight predictors, which could more easily be persisted, on which we could maybe one day ensure compatibility across versions...

This is easy to do for linear models, very hard in the broad sense, but we could do things progressively, and possibly not fully.


In short, if we could make the "freeze" method a bit more ambitious in the long term, I see a value in having it as a method, and I would be enthusiastic.

PR #101, opened 2026-03-17
SLEP026: Add ``scoring`` to the ``score`` Method and Remove the Default

lorentzenchr:


slep026/proposal.rst
```diff
@@ -0,0 +1,350 @@
...
+- Very likely a major release for something not very marketable (a deprecation).
+- More imports required and a bit longer code.
+  Instead of::
+  
+      >>> estimator.score(X, y)
+      
+  one needs::
+  
+      >>> from sklearn.metrics import super_metric
+      >>> super_metric(y_true=y, y_pred=estimator.predict(X))
```
thomasjpfan: I think most of the pain comes from having to know which form of predict a metric needs: (`predict`, `predict_proba`, or `decision_boundary`). For some metrics, you may need to slice `predict_proba`: `est.predict_proba(X)[:, 1]`.

Two benefits of `estimator.score(X, y, scoring="r2_score")`:
- Do not have to worry about the shape of the prediction.
- If one is passing in `scoring` to a `GridSearchCV(est, scoring="...")`, they can run `est.score(..., scoring="...")` to see how the metrics behave before doing the full grid search. The API is a little more consistent.

slep026/proposal.rst
```diff
@@ -0,0 +1,350 @@
...
+In principle, this deprecation can be accomplished with the usual deprecation policy
+of scikit-learn. It is, however, not backward compatible in the sense that, as with all
+deprecations, future versions won't work with code written for older versions: an
+``AttributeError`` is raised if ``estimator.score(X, y)`` is called, a
+``ValueError`` is raised if ``cv_tool(scoring=None)`` is called, a ``TypeError`` for
+``cv_tool()`` (because ``scoring`` has no default value anymore).
+
+4.3. Downstream packages
+------------------------
+Many packages inherit from ``ClassifierMixin`` and ``RegressorMixin``, some examples include
```
thomasjpfan, line 282: This is a good point. We can go through a transition for scikit-learn estimators, but impacting third party estimators feels a bit more painful.

slep026/proposal.rst
lorentzenchr: > Do not have to worry about the shape of the prediction

If that’s the case, shouldn’t we fix those metrics/scores? Do you have an example?

I don’t understand you 2nd bullet.

slep026/proposal.rst
lorentzenchr: XGBoost overrides the score method, so no problem there.
LightGBM just inherits them. They would also inherit the deprecation and removal. They don’t need to do much if anything.

thomasjpfan: I'm mostly thinking about if LightGBM wants to keep `score`, they would need to implement it themselves. But they will find out that does not work anymore for first party scikit-learn meta-estimators.

There could also be third party meta-estimators (Like a `SuperCustomSearchCV`) that depend on `.score`, but also need to push users to pass in a `scoring="some-metric-string"`.

slep026/proposal.rst
thomasjpfan: > If that’s the case, shouldn’t we fix those metrics/scores? Do you have an example?

An example is `roc_auc_score`, which can run with:

```python
roc_auc_score(y_true, est.predict_proba(X)[:, 1])
```

or

```python
roc_auc_score(y_true, est.decision_function(X))
```

Users need to understand the shape of the prediction to properly handle the metric. I guess the workaround is to:

```python
scorer = get_scorer("roc_auc")
result = scorer(est, X, y)
```

For the grid search example, it's the same concern. I'm thinking of someone that is using `GridSearchCV(..., scoring="roc_auc")` and wonder "how do I manually compute the score?". If they use `get_scorer`, they they do not need to learn about the shape of the input to the underlying metric.

lorentzenchr: Yes, `get_scorer` was also on my mind.
I guess we should just fix the `roc_auc_score` with `predict_proba`.

lorentzenchr:
@thomasjpfan @adrinjalali @glemaitre @ogrisel ping. As discusses yesterday during the sprint.

slep026/proposal.rst
```diff
@@ -0,0 +1,397 @@
...
+While accuracy might be the simplest score to explain, R2 is already not very intuitive
+and, evaluated out-of-sample (not on the training set), it even lacks its simple
+interpretation in terms of variances.
+
+**Conclusion** as stated above: *There is no good default score.*
+
+3.2 (Mis-) Use cases
+--------------------
+
+1. Using R2 for quantile regression. Note that ``QuantileRegressor.score`` computes R2.
```
adrinjalali, line 91: would be nice to hint at why this is wrong

slep026/proposal.rst
```diff
@@ -0,0 +1,397 @@
...
+``cross_val_score``         ``scoring``     default ``None``  ``None`` means estimator's ``score`` method
+``cross_validate``          ``scoring``     default ``None``  ``None`` means estimator's ``score`` method
+``permutation_test_score``  ``scoring``     default ``None``  ``None`` means estimator's ``score`` method
+``validation_curve``        ``scoring``     default ``None``  ``None`` means estimator's ``score`` method
+``LearningCurveDisplay``    ``score_name``  default ``None``  ``None`` means estimator's ``score`` method
+``ValidationCurveDisplay``  ``score_name``  default ``None``  ``None`` means estimator's ``score`` method
+==========================  ==============  ================  ===========================================
+
+==========================  ==========================  ================  ===========================================
+meta data routing           method/attribute/parameter  score used        comment
```
adrinjalali: not sure what metadata routing has to do here

slep026/proposal.rst
```diff
@@ -0,0 +1,397 @@
...
+4. Solution
+===========
+4.1. Add ``scoring`` to ``score`` and remove the default
+--------------------------------------------------------
+**The proposed solution is:**
+
+- Add a ``scoring`` parameter to ``score`` such that a user can select the used score
+  as it is already the case in different classes and functions.
+  Keep the current default, now set explicitly.
+- Deprecate and remove the default from the ``scoring`, for supervised classifiers,
```
adrinjalali, line 248: I suspect removing this point, and simply adding a constant user warning w/o deprecation, would make merging this slep easier. I personally would be happy as long as we do the warning.

slep026/proposal.rst
lorentzenchr: added

slep026/proposal.rst
lorentzenchr: removed

PR #104, opened 2026-04-03
FIX Include estimator in the hook signatures

jeremiedbb:
The `estimator` that called the hook was previously accessible as an attribute of the context, but we realized that it is not a good idea because we really don't want to pickle the estimator when it's not necessary. For instance:
- when a callback sends the context to the callback manager and only needs the task info from the context (e.g. `ProgressBar`)
- when a meta-estimator fits its sub-estimator in parallel, the context and therefore its ancestors are sent to the workers but the meta-estimator is not needed in the workers (e.g. `*SearchCV`)

(ref https://github.com/scikit-learn/scikit-learn/pull/33664 for more details)


slep023/proposal.rst
```diff
@@ -162,10 +162,10 @@ Callbacks must implement the following
 .. code-block:: python
 
     class FitCallback(Protocol):
-        def setup(self, context): ...
-        def on_fit_task_begin(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
-        def on_fit_task_end(self, context, *, X=None, y=None, metadata=None, fitted_estimator=None): ...
-        def teardown(self, context): ...
+        def setup(self, estimator, context): ...
```
thomasjpfan, line 165: Would `FitCallback` store `estimator` (i.e. `self._estimator = estimator`)? If so, that would be a circular reference.

jeremiedbb: no it doesn't store it. I mean we can't prevent third party callbacks to do it, but sklearn built-in callbacks won't.

adrinjalali (approved):

